PR #3, opened 2024-01-30
Send functions lua

PMassicotte:
- Also started code to detect code chunks (r, python)


jalvesaq:
In a couple of hours, I will commit changes converting a lot of VimScript to Lua.

PMassicotte:
Perfect, I will then sync and see where I can continue to work!

PR #5, opened 2024-02-01
Selene stylua vim

she3o:


she3o:
This PR is ready for review. It just adds to CONTRIBUTING.md and adds stylua to the git hook. the pre-commit hook is severely refactored.

jalvesaq:
Any style standard is better than none. I have this in my `init.lua`:

```lua
vim.o.expandtab = true
vim.o.shiftwidth = 4
```

But I can switch to `expandtab = false` and turn it `true` only for Python.

@PMassicotte ?

PMassicotte:
I have shiftwidth set to 2.

I also use neoformat for lua with these options:

```lua
    vim.g.neoformat_lua_stylua = {
      exe = 'stylua',
      args = {
        '--call-parentheses Always',
        '--indent-type Spaces',
        '--quote-style AutoPreferSingle',
        '--indent-width 2',
        '--column-width 80',
      },
      replace = true,
    }
```

PMassicotte:
I can change to whatever you think is preferable.

jalvesaq:
Thanks for your flexibility, @PMassicotte!
Luaformater was updated for the last time 9 years ago: https://github.com/LuaDevelopmentTools/luaformatter
Is that correct? I have just tried Stylua and the result is similar to what I'm writing without any formatting tool other than the default Neovim indentation. We really need to follow a standard to make resolving git conflicts easier. So, I'm going to merge this pull request now.

jalvesaq:
Is there a way of preventing stylua of inserting so many line breaks? For me, it's easier to read the code if I can see more statements in a single glance. Sometimes, I also manually aligned table elements, as in the `config` table in the `config.lua` and stylua misaligned them. Is there a less aggressive formatter?

jalvesaq:
I'm learning that the line breaks depend on the `column_width` parameter...

But I still prefer this:

```lua
    if file_type == "quarto" then
        create_maps('n', 'RQuartoRender',  'qr', ':lua require("r.quarto").command("render")')
        create_maps('n', 'RQuartoPreview', 'qp', ':lua require("r.quarto").command("preview")')
        create_maps('n', 'RQuartoStop',    'qs', ':lua require("r.quarto").command("stop")')
    end
```
than

```lua
    if file_type == "quarto" then
        create_maps("n", "RQuartoRender", "qr", ':lua require("r.quarto").command("render")')
        create_maps("n", "RQuartoPreview", "qp", ':lua require("r.quarto").command("preview")')
        create_maps("n", "RQuartoStop", "qs", ':lua require("r.quarto").command("stop")')
    end
```

PMassicotte:
Do you mean using single vs double quotes (in the last example)?

jalvesaq:
No, it was the vertical alignment of arguments. I slightly prefer double quotes because in C double quotes are used for strings and single quotes for single characters, and part of our code is in C, but I would be fine with single quotes too.

I merged your pull request. Thanks!

Now, I will format the whole code with stylua, using `ignore` comments to keep the vertical alignment at some places.
I prefer a width of 80 columns to be easier to vertically compare files for changes. So, I will try 90 columns...

We can change the style later. Perhaps the problem is that I still have to learn how to use git properly...

PR #9, opened 2024-02-02
Port nvimrserver refactor

she3o:


jalvesaq:
I updated the names of callback functions from Vim to Lua in nvimrserver.c, deleted one function, and put `void` in some function declarations to avoid `clangd` warnings.

jalvesaq:
I fixed a bug: you declared listTree twice. See: https://github.com/jalvesaq/tmp-R-Nvim/commit/98937aec0fdd84af1df55273511a74b9e3b355a0

PR #13, opened 2024-02-04
Update config.lua

PMassicotte:
change auto_start config to be a boolean (like other params)


jalvesaq:
It's not a boolean. There are three valid values.

PMassicotte:
Ah ok I missed it.

PR #15, opened 2024-02-06
Vim to lua func

PMassicotte:


jalvesaq:
Thanks! I will try to convert the remaining of R/start_r.vim today. You can refactor the send functions later.

PR #16, opened 2024-02-06
Luacheck action

PMassicotte:
This adds a github action using luacheck


PMassicotte:
I will close this as I am working on a better way using selene

PR #17, opened 2024-02-07
Github actions using the selene linter

PMassicotte:
Every time that code is pushed, the linter will be executed. Checks will fail if code structure is not respected. We can use it to standardize all the Lua code in the project.


jalvesaq:
I have no experience with Github actions. So, perhaps @she3o may comment on this pull request...

I'm running `selene` manually. It says that the "goto" statement in `lua/r/job.lua` is an error, but it works.

I will have limited time to work on this project in the next few days.

jalvesaq:
Thanks! I will merge to see how it works.

PR #18, opened 2024-02-08
Use math.min and math.max in doc.lua

she3o:


jalvesaq:
This doesn't seem to be equivalent:

```diff
-            htwf = (wwidth > 80) and 88.1 or ((wwidth - 1) / 0.9)
+            htwf = math.min(wwidth, 80)
```

jalvesaq:
Perhaps we should simplify how the help window is open to the same algorithm used by Vim for its `:help`:

```
:h[elp]	Open a window and display the help file in read-only
	mode.  If there is a help window open already, use
	that one.  Otherwise, if the current window uses the
	full width of the screen or is at least 80 characters
	wide, the help window will appear just above the
	current window.  Otherwise the new window is put at
	the very top.
```

PR #19, opened 2024-02-08
Cleaner config.lua

she3o:


jalvesaq:
Github doesn't allow the merge due to selene finding errors.

jalvesaq:
I checked your branch out. It has loop errors. This means that module X requires module Y, which requires module X. I'm fixing this kind of issue by requiring module Y inside the function that uses it, not at the top of module X.

jalvesaq:
Could you move the line `local edit = require("r.edit")` from the top of `config.lua` to the function where it is used?

she3o:
I'll soon finish my work shift and be on it (Can it be done from Github mobile?).

Maybe I should just revert the second commit where I define `r.edit` and replace its references. I expect a lot of the functions will be moved around files anyway. Does that seem right?

jalvesaq:
> I'll soon finish my work shift and be on it (Can it be done from Github mobile?).

I never used Github mobile, but there is no hurry.

> Maybe I should just revert the second commit where I define `r.edit` and replace its references.

I think this would be enough for now.

> I expect a lot of the functions will be moved around files anyway. Does that seem right?

Yes, at some point we may think about moving functions to more appropriate modules. Other tasks are:

  - Do `grep -R '\\|' lua/` and fix patterns that don't work.

  - Document functions, or, at least, add  the `@param` to allow linters to find functions called with invalid parameters.

jalvesaq:
It seems tha `uv.fs_access` requires the file's full path. So, it returns `false` when checking if "R" is executable.

jalvesaq:
As far as I know, Lua doesn't support patterns with `|`.

I will fix these issues in the next commit.

PR #24, opened 2024-02-10
Document jobs.lua

she3o:


jalvesaq:
Using the command `:Neogen`, I noted this difference:

with `-- @param`:

image

with `---@param`:

image

Only with the second format does the LSP catch errors of parameters with wrong types.

PR #25, opened 2024-02-10
Replace `NVIMR_` with `RNVIM`, `R-Nvim` with `R.nvim`, and `nvimrserver` with `rnvimserver`

jalvesaq:
Since the plugin name will be `R.nvim`, I think we should change the prefix of environment variables accordingly.


PMassicotte:
Looks all good to me!

jalvesaq:
I did more replacements... and I think we should also rename `nvimrserver` as rnvimserver`.

jalvesaq:
I suppose it's finished now. I can't see anything else to rename.

PR #28, opened 2024-02-10
Remvoe the echo parameter

PMassicotte:
refactor(maps.lua): remove unnecessary file_type parameter from send function
refactor(maps.lua): simplify create_maps calls by removing unnecessary 'silent' and 'echo' arguments
refactor(rmd.lua): remove unused 'e' parameter from send_R_chunk function
refactor(rnw.lua): remove unused 'e' parameter from send_chunk function
refactor(send.lua): remove unused 'e' parameter from marked_block and selection functions

I think we will have to also remove the `M.get_source_args` in `source_file` and only call it once just before sending the code to the terminal.

### DIFF
--- a/lua/r/maps.lua
+++ b/lua/r/maps.lua
@@ -130,29 +130,29 @@ end
 
 local send = function(file_type)
     -- Block
-    create_maps("ni",  "RSendMBlock",      "bb", "<Cmd>lua require('r.send').marked_block('stay')")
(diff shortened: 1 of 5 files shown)


jalvesaq:
Since "m" has only two possible values, it could be a boolean: true = move, false = stay put.

PMassicotte:
I think this should be this!

PMassicotte:
Realized that the send paragraph function was missing the first line of the paragraph (0 based).

PR #30, opened 2024-02-12
Add the ability to send part of the curent pipe chain to the R console/terminal

PMassicotte:
The idea is based on breakerofchain (https://github.com/MilesMcBain/breakerofchains) and aims at sending part of the current piping chain in the terminal. The goal is to facilitate debugging and see intermediate results.

Peek 2024-02-12 14-04


jalvesaq:
Very nice! And it works with `ggplot` too (`+` as pipe operator).

PMassicotte:
> Very nice! And it works with `ggplot` too (`+` as pipe operator).

yes, I use that alot when making graphs :)

PR #31, opened 2024-02-13
Fix sending above lines

PMassicotte:
fix(send.lua): fix above_lines function to include the current line in the range of lines to send to R…n the range of lines to send to R

### DIFF
--- a/lua/r/send.lua
+++ b/lua/r/send.lua
@@ -147,7 +147,7 @@ end
 
 --- Send to R all lines above the current one.
 M.above_lines = function()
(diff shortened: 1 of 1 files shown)


PMassicotte:
fully transfered to tmp-R-Nvim and all work fine now!

jalvesaq:
Great! We still have to do some code optimizations, which can introduce new bugs. I prefer to inaugurate R.nvim when we have only new features to introduce.

PMassicotte:
Completely agree. Not sure if a placebo effect, but I have the feeling it is now much more snappy (completion, sending lines, ...).

PR #33, opened 2024-02-15
Use M.source_lines instead of M.cmd so we can send lines that contains function call spread over multiple lines

PMassicotte:
For example, sending these lines (using send above) was not working:

```r
library(tidyverse)

mtcars |>
  head() |>
  mutate(
    drat = drat / 2, hp
  ) |>
  mutate(y = 1 + 1)
```

Because the `mutate()` call was spread over 3 lines, and each of these line were sent separately to the terminal.


jalvesaq:
Thanks!

PR #35, opened 2024-02-16
Better handling of the chain break function

PMassicotte:
This is taking care of edge cases where the code in a pipe chain could cover over multiple likes.

For example, this was not working (fixed now).

```r
mtcars |>
  mutate(
    y =1
)
```


jalvesaq:
The function `trim_lines` is global. It should be local.

jalvesaq:
Is it ready to merge?

PMassicotte:
Yup

PR #38, opened 2024-02-19
Use treesitter to source functions!

PMassicotte:
This adds 2 new mapping:

`<LocalLeader>fc` to send the current function
`<LocalLeader>fa` to send all the functions in the current buffer

There are no extra dependencies to install (works only with neovim, not vim).

feat(maps.lua): add new mappings for sending all functions in the current buffer
feat(maps.lua): add new mappings for sending the current function where the cursor is
feat(send.lua): implement function to send all functions in the current buffer
feat(send.lua): implement function to send the current function where the cursor is

Demo

Peek 2024-02-19 15-42


jalvesaq:
This is great! I was hoping that someone would do this in the future, but you already did it! I will try it later today.

PMassicotte:
The only canvas I have to fix is that nested functions will be also sent and sourced

````r
myfun <- function(x) {
  x + 1


  x + 4
}

myfun2 <- function(x) {
  x + 1

  name <- function(variables) {

  }
}


x <- 1
```

jalvesaq:
As I understand, a nested function is part of the nesting function and naturally should be sourced as ordinary lines are.

Imagine that you have a function with 10 lines. Even if you change only 2 lines, you have to send the whole function to R Console again.

PMassicotte:
In my code snippet, the _name_ function will appear in the object browser.

jalvesaq:
Probably, you are sending all lines separately. The function `send.cmd()` concatenates all lines before sending them as a single command with line breaks within. See `send_cmd_to_term()` in `term.lua`.

jalvesaq:
What I said doesn't make sense. `source_lines()` also concatenates the lines before sending them.

PMassicotte:
The issue is that my treesitter code captures all functions, I am changing that to skip inner functions.

PMassicotte:
All fixed now!

PMassicotte:
I have to go for now, but looks like it is working fine. If you find issues, let me know, I will fix that tomorrow.

jalvesaq:
The only problem that I noted is that it isn't getting a function defined with `=` as the assignment operator.

jalvesaq:
Another issue is that it should work if in a block of R code within an RMarkdown or Quarto document.

The tree-sitter parser for Rnoweb doesn't recognize R code, and there is no tree-sitter parser for Rhelp.

PMassicotte:
Good I will work on that today.

jalvesaq:
If I disable all plugins with "treesitter" in their names, I get this error:

```
...local/share/nvim/runtime/lua/vim/treesitter/language.lua:102: no parser for 'r' language, see :help treesitter-parsers
```

The error comes from `r_fun_query`. So, it seems that we have to check if the tree-sitter parser for "r" is installed before calling `vim.treesitter.query.parse()`.

PMassicotte:
> If I disable all plugins with "treesitter" in their names, I get this error:
> 
> ```
> ...local/share/nvim/runtime/lua/vim/treesitter/language.lua:102: no parser for 'r' language, see :help treesitter-parsers
> ```
> 
> The error comes from `r_fun_query`. So, it seems that we have to check if the tree-sitter parser for "r" is installed before calling `vim.treesitter.query.parse()`.

One can call `TSInstall r`. I should try on vanilla nvim to see if it is sufficient.

jalvesaq:
Perhaps we could insert all lines of all functions in a single `string[]` table and call `M.source_lines()` after the loop `for`:

```lua
M.source_lines(all_lines, "function")
```

The `what` argument of `source_lines()` is a cosmetic artifact to let insecure users know that what is being sourced is a function, paragraph, marked block, or selection of lines.

jalvesaq:
`:TSInstall` is a command of the `nvim-treesitter` plugin.

PMassicotte:
After a bit of digging, I found out that the treesitters aren't yet available for the rmd and qmd files. Hence, sending functions will only work for R files. I have added code what will notify the user if they try to use it in either of these file types.

jalvesaq:
Did you try the parses "markdown" and "markdown_inline"?

PMassicotte:
> Did you try the parses "markdown" and "markdown_inline"?

Yes, but R nodes are not defined in markdown. 

image

We see them in the syntax tree, but only because treesitter uses injection.

PMassicotte:
rmd and qmd are currently considered as simple markdown by tree-sitter.

jalvesaq:
Note: you have to close this pull request and do it again because I disordered git history yesterday.

Issue #43, opened 2024-02-21
Could not find R path in Windows Registry


akthe-at:
On Windows 10, as of this morning with the new updates over the last 12 hours I am not able to build/install nvimcom. During the installation process I get an error that says Could find R path in Windows Registry. If you already installed R, please, set the value of 'R_path'. 

I am wondering if I missed somewhere that the config options from before need to be altered for the new R-nvim, specifically items like (vim.g.R_path='')

akthe-at:
Everything still works at least.

Further testing found this as the only ways to get around the "error" message so far. Either by A, using bash to start Neovim or by opening a file directly with Neovim aka "nvim main.r" from PowerShell. It doesn't seem to matter how i change commands like R_path it doesn't impact the message. 

Overall, the message does not seem to impact functionality at all.

akthe-at:
I am using Git Bash technically, run from inside the Wezterm Terminal (Git Bash downloaded as a portable standalone binary). :echo has(“win32”) = 1 anf Sys.info() has sysname = “Windows”.

The above holds true from Neovim started inside a Bash shell or Powershell…shell.

Adam Kelly, MS
Data Reporting & Analysis Specialist │ BerbeeWalsh Department of Emergency Medicine
University of Wisconsin-Madison School of Medicine & Public Health
800 University Bay Drive Suite 310, Mail Code 9123, Madison, WI 53705

PR #45, opened 2024-02-22
allows proper R_path lookup in registry

akthe-at:
Fixes issue [#43](https://github.com/R-nvim/R.nvim/issues/43) from yesterday! Error message goes away and doesn't seem impact any other functionality.


jalvesaq:
Thanks!

PR #46, opened 2024-02-23
Behave as documented

jalvesaq:
- Send marked block: don't include line of second mark.
- Open html according to options (0, 1, 2).
- Go to next/previous chunk v:count1 times.

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -264,12 +264,12 @@ that follows <LocalLeader>m, e.g. <LocalLeader>m} (from the current line to
 the end of paragraph) or <LocalLeader>m3j (the next 3 lines including the
 current one).
 
-If the cursor is between manually inserted marks, the plugin will send the
-lines between them to R if you press <LocalLeader>bb. If the cursor is above
-the first mark, the plugin will send from the beginning of the file to the
-mark. If the cursor is below the last mark, the plugin will send from the mark
-to the end of the file. The mark above the cursor is included and the mark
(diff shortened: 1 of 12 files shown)


jalvesaq:
I revised R.nvim documentation and noted three features that were not implemented as they were in Nvim-R:

1. When sending a marked block to R, the line of the second mark should not be included. It was defined this way in Nvim-R to allow users to mark various lines in an R script and send one block after the other. If both, the first and the second marks had their lines included, each marked line would be sent twice. I still would like to find an efficient way of replacing `vim.fn.line("$")` for which `lua_ls` warns that the result might `nil`.

2. Opening an HTML was not being like a PDF, although the options are the same (0: don't open; 1: open, but don't focus; 2: open and focus). I fixed the implementation and added a note to the documentation explaining that in many cases R.nvim will not be able to do what is expected. I moved the function that focuses the window from `pdf/init.lua` to `utils.lua` because now it's also used to focus a browser window.

3. When going to a previous/next chunk of code (Rmd, Quarto, or Rnoweb), the variable `v:count` must be considered. For example, 3\gn means "jump to the third chunk below". This happened automatically in VimScript, but I had to code it manually in Lua.

jalvesaq:
I believe this pull request is ready to be merged, but, of course, it would be better if someone could review it.

PMassicotte (approved):
I have tested the branch and everything seems to work on my side.

jalvesaq:
Thank you! I will merge it.

PR #55, opened 2024-02-25
Convert some options from number to string

### LINKED ISSUE #52: Better `auto_start` option?
Currently, `auto_start` options are 0, 1, and 2, but users will only know what these numbers mean if they remember what is in the documentation. So would it be better to change the option type from number to string? If yes, what about "no", "on startup", and "always" as valid values?

jalvesaq:
`auto_start`: "no", "on startup", "always"
`open_html` and `open_pdf`: "no", "open", "open and focus" `nvim_wd` becomes `setwd`: "no", "file", "nvim"
Close #52


jalvesaq:
In addition to converting `auto_start` to string, following the same idea, I'm also suggesting the conversion of other options from number to string.

PMassicotte (approved):
Good job! Sounds perfect to me!

PR #58, opened 2024-02-26
Add annotations to some functions

jalvesaq:
- Add annotations to functions that either receive parameters or return any value and fix the consequent new warnings from lua_ls.
- Delete unused function, browser.print_list_tree(), and the corresponding code in rnvimserver.

### DIFF
--- a/lua/r/browser.lua
+++ b/lua/r/browser.lua
@@ -21,6 +21,7 @@ local hasbrowsermenu = false
 --- Escape with backticks invalid R names
 ---@param word string
 ---@param esc_reserved boolean
+---@return string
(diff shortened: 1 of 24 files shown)


PMassicotte:
Very  nice! Good documentation is so important. Are you using a tool to get a template for each function?

jalvesaq:
> Are you using a tool to get a template for each function?

I'm using the `:Neogen` command from [danymat/neogen](https://github.com/danymat/neogen).

PMassicotte:
Tx for sharing, I will get that. Anyway, you did a good job on documentation!

PR #50, opened 2024-02-23
Modify test setup to allow testing in live Neovim sessions. 

she3o:
- Introduce Busted test configuration `.busted` for Lua tests, specifying custom test runner.
- Add Makefile with targets for running tests, showing help, and placeholder for coverage.
- Implement testing script `tests/run` as a Lua interpreter wrapper for Neovim.
- Adjust `tests/init.lua` and `rmd_spec.lua` for new testing infrastructure.

This commit will allow us to test the plugin in a live Neovim session. It also includes TODOs and `pending()` for future implementation.

PR #60, opened 2024-02-29
Replace vim.toml with vim.yml

she3o:
Which makes it more readable. 

I have also defined 'pending' which will hopefully make selene tolerate PR #50


jalvesaq:
I didn't note any problem caused by the `vim.yml`. So, I think we merge this.

PR #63, opened 2024-03-01
Fix some functions

she3o:


jalvesaq:
I missed the `nvim_buf_line_count()`... Thanks for finding it!
Since now `get_last_line_num()` in `r.utils` is a single line, we can delete it and call `vim.api.nvim_buf_line_count(0)` directly instead.

The motion function is not being called. I found that calling a Lua function with `v:lua.` must use single quotes without parentheses (see `:h v:lua-call`). Then, in `lua/r/maps.lua`, the mapping should change to:

```lua
    create_maps("ni.", "RSendMotion",         "m",        "<Cmd>set opfunc=v:lua.require'r.send'.motion<CR>g@")
```

But even after this, the motion doesn't seem to be correctly recognized.

PR #74, opened 2024-03-08
Remove completion based on class of first object

jalvesaq:
Close https://github.com/R-nvim/cmp-r/issues/6

### DIFF
--- a/nvimcom/DESCRIPTION
+++ b/nvimcom/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: nvimcom
-Version: 0.9.28
-Date: 2024-03-06
+Version: 0.9.29
+Date: 2024-03-08
 Title: Intermediate the Communication Between R and Neovim
 Author: Jakson Aquino
 Maintainer: Jakson Alves de Aquino <jalvesaq@gmail.com>
--- a/nvimcom/R/misc.R
+++ b/nvimcom/R/misc.R
@@ -423,13 +423,6 @@ nvim.getclass <- function(x) {
     return(cls)
 }
 
-nvim.getmethod <- function(fname, objclass) {
-    mname <- paste0(fname, objclass)
-    if (exists(mname) && is.function(get(mname)))
-        return(mname)
-    return(fname)
-}
-
 #' Complete arguments of functions.
 #' Called during nvim-cmp completion with cmp-r as source.
 #' @param id Completion identification number.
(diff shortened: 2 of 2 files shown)


jalvesaq:
@PMassicotte, this pull request fixes https://github.com/R-nvim/cmp-r/issues/6 by removing the feature, as discussed on the issue page. So far, the only way of completing function arguments based on the class of the first argument would be to parse the documentation. But R itself displays complete and correct options for completion when we press TAB in the R Console. So there must be a way of doing it.

PMassicotte:
What feature it will remove exactly? Can you show an example?

PMassicotte:
Could we use something like `attr(methods("print"), "info")` ?

jalvesaq:
It is what I've described at https://github.com/R-nvim/cmp-r/issues/6#issuecomment-1984339786 :

Type these commands in the R Console (or radian):

```r
library(tidyverse)
mt <- tibble(mtcars)
print(mt, 
```

Then press `<Tab>` and wait about 2 seconds for the completion options to be displayed. The arguments `max_extra_cols` and `max_footer_lines` (from the method `print.tbl`) will be among the options.

PMassicotte:
If I understand, after this PR, `max_extra_cols` won't show anymore?

PMassicotte:
image

This works for S3 and S4

PMassicotte:
Maybe we can merge this for now and explore `attr(methods("print"), "info")` later.

jalvesaq:
As far as I remember, years ago, S3 methods were defined implicitly: any function with a dot in its name would be a candidate to be a method. For example, if you defined a function called `print.factor` it would be used as the print method for objects of class `factor`. The completion with `cmp_r` was supposed to work for this old case, but even in this case, there is a bug. I will close this pull request, fix the bug, and, then, we can think about what to do next.

PMassicotte:
Ok cool we can look at this later.

Some pointer using the `sloop` R package (https://sloop.r-lib.org/):

```r
r$> library(sloop)

r$> s3_dispatch(print(tibble::tibble()))
   print.tbl_df
=> print.tbl
 * print.data.frame
 * print.default
```

```r
r$> s3_methods_class("tbl")
# A tibble: 2 × 4
  generic class visible source             
  <chr>   <chr> <lgl>   <chr>              
1 format  tbl   FALSE   registered S3method
2 print   tbl   FALSE   registered S3method
```

jalvesaq:
The command `methods(print)` lists a lot of methods. Most of them have an `*` at the end of their names and `R.nvim` completion works only with the few mthods that don't have the `*`. Example:

```r
# The generic `print()` and the implicitly defined methods `print.factor()` and
# `print.table()` receive different arguments:
args(print)
args(print.factor)
args(print.table)

# R.nvim can correctly complete the arguments for `print.factor` and `print.table`:
fctr1 <- factor(c("a", "b", "a", "b", "a"))
fctr2 <- factor(c("c", "d", "d", "c", "d"))
tbl <- table(fctr1, fctr2)
print(fctr1, )
print(tbl,  )

# But not of `print.lm*`
mdl <- lm(rnorm(100) ~ runif(100))
print(mdl, )

# If we do `\rh` over the word `print` in `print(mdl, )` R.nvim will open the
# documentation of the `print.lm*` method.
```

PMassicotte:
For the `print(mdl,  )`  I am getting this:

image

Is this expected or should I get a different set of arguments?

jalvesaq:
Maybe `print.lm` wasn't a good example because it doesn't have specific arguments.

`print.summary.lm` is better: `symbolic.cor` and `signif.stars` should be among the arguments in the completion menu:

```r
mdl <- lm(rnorm(100) ~ runif(100))
sm <- summary(mdl)
print(sm, )
```

PR #78, opened 2024-03-09
Complete methods arguments

jalvesaq:


PMassicotte:
Great working now. Not sure what was the issue with github. I am now getting the correct completion for `print.summary.lm`. I even tested `<localleader>rh` on some of my packages implementing `print` and it work perfectly.

PMassicotte:
How the `::` works exactly? If the library was in the global environment, it was already working. Should it work now with any lib (i.e. not just those in the globenv)?

jalvesaq:
For `print.summary.lm` I get the documentation for `symbolic.cor`, but not for `signif.stars`. The reason is that `signif.stars` isn't listed as argument for `summary.lm` in `~/.cache/R.nvim/args_stats_4.3.3`.

jalvesaq:
> How the `::` works exactly? If the library was in the global environment, it was already working. Should it work now with any lib (i.e. not just those in the globenv)?

Only those in `.GlobalEnv`. We can add support for non-loaded packages, but this would be a general change, not specific to this pull request.

PMassicotte:
You mean for the documentation on the complete menu? Yes, I also have it. I think that is already a big improvement!

PMassicotte:
Anyhow, you did a very good job on this!

jalvesaq:
Thank you! I will see if there is anything easy to do about the documentation of items.

PMassicotte:
Could we do something like `help(print.summary.lm, help_type = "text")` and parse the results?

jalvesaq:
Currently, the `::` are useful only to complete arguments from a masked function. Example:

```r
dplyr::filter()
stats::filter()
```

jalvesaq:
> Could we do something like `help(print.summary.lm, help_type = "text")` and parse the results?

It would be a lot of work, but it's an option if my idea doesn't work. I will see if only items from the visible function are being saved in the `args_` file. If yes, then, I will see if it's easy to save all items.

PMassicotte:
Can also save it to html with `help_type`. That could make the parsing much easier.

jalvesaq:
> Can also save it to html with `help_type`. That could make the parsing much easier.

I've tried this in the past. It's not easier because in Rd we have curly brackets delimiting sections, items within sections, etc.

`help_type` is plain text, right? In addition to being more difficult to identify sections, it's more difficult to distinguish between code and links, etc...

I didn't try HTML, though.

jalvesaq:
The problem is here: https://github.com/R-nvim/R.nvim/blob/f8ab041151532088751145c3066a0ca93a619980/nvimcom/R/bol.R#L522

We request only the documentation for arguments of the main documented function (`frm` are the `formals()` from the function). We should get all arguments available in the documentation. It will be harmless because they are displayed only when the corresponding item is selected.

PMassicotte:
That seems to be the easiest way at the moment.

PR #82, opened 2024-03-11
Improve completion of methods arguments

jalvesaq:
- Look for methods for second and following classes in objects with multiple classes.
- Always run code to build completion data as a Neovim job.
- Build the `args_` file right after building the corresponding `objls_` file.
- Add time to build `objls_` + `args_` to `:RDebugInfo`.

### DIFF
--- a/nvimcom/R/misc.R
+++ b/nvimcom/R/misc.R
@@ -427,26 +427,28 @@ nvim.getmethod <- function(fname, objclass) {
     if (exists(fname) && is.function(get(fname)) &&
         (isGeneric(fname) || isS3stdGeneric(fname))) {
         mtd <- rownames(attr(methods(fname), "info"))
-        fnm <- paste0(fname, ".", objclass)
-        idx <- grep(paste0("^", fnm, "$"), mtd)
-        if(length(idx) == 1) {
-            fun <- NULL
-            try(fun <- getS3method(fname, objclass))
-            if (is.function(fun)) {
-                frm <- formals(fun)
-                luatbl <- sapply(frm,
-                                 function(x)
-                                     if (length(x) == 0) {
-                                         return("")
-                                     } else {
-                                         return(" = ")
-                                     })
-                env <- paste0(fname, "\x02", fnm)
-                clpstr <- paste0( "', cls = 'm', env = '", env, "'}, {label = '")
-                luastr <- paste0(names(luatbl), unname(luatbl),
-                                 collapse = clpstr)
-                luastr <- paste0("{{label = '", luastr, "', cls = 'm', env = '", env, "'}}")
-                return(luastr)
+        for (obc in objclass) {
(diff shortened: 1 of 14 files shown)


jalvesaq:
This pull request should be tried with the branch `improve_methods_args` from `cmp-r`.

Building an `args_` file was about three times slower than building the corresponding `objls_` file. After some optimization, it's now about 5 times faster than before. The big improvement in the performance was due to avoiding calling `help()` just to get the directory where the package was installed. Now that it's much faster, it is being built right after the `objls_` file. This simplifies the code, which now has about 260 fewer lines.

PMassicotte:
oh let me try it!

PMassicotte:
image

Thas is nice!

PMassicotte:
Are we missing some parameters?

image

jalvesaq:
`correlation` is for `summary.lm`, not for `print.summary.lm`

PMassicotte:
oh wow, I still need coffee this morning :)

I have to make some change to a project, I will report back in 1 h ( I will use your PR)

jalvesaq:
In this example, the R documentation is shared by two functions and the `Arguments` section combine all arguments of the two functions.

PMassicotte:
This feels very fast already

jalvesaq:
The completion by `rnvimserver` + `cmp-r` is almost instantaneous. The only slowness is making the cache files when the library is loaded for the first time.

nvimcom/R/misc.R
```diff
@@ -427,26 +427,28 @@ nvim.getmethod <- function(fname, objclass) {
...
-                                     } else {
-                                         return(" = ")
-                                     })
-                env <- paste0(fname, "\x02", fnm)
-                clpstr <- paste0( "', cls = 'm', env = '", env, "'}, {label = '")
-                luastr <- paste0(names(luatbl), unname(luatbl),
-                                 collapse = clpstr)
-                luastr <- paste0("{{label = '", luastr, "', cls = 'm', env = '", env, "'}}")
-                return(luastr)
+        for (obc in objclass) {
```
PMassicotte, line 430: Nicely done!

PMassicotte:
Well, I have worked roughly 1 hour and it works perfectly fine.  Speed is not comparable as it was before. You did a very nice job here :) I do not see any problem at mergin this :+1:

jalvesaq:
I ran it with Valgrind and there was no invalid read or write in the C code. So, I believe that it's safe to merge. If any bugs appear, we will fix them.

PMassicotte:
I have tried it also on a Window machine and no issue.

jalvesaq:
Thank you!

PMassicotte:
Just need to merge the cmp branch

jalvesaq:
I'll do it... Thanks for remembering!

PR #99, opened 2024-03-25
Send some preceding lines to the console

wurli:
For example, hitting `<CR>` with the cursor on the second line here, all 3 lines will now be sent to the console:

``` r
iris |>
  subset(Species == "setosa") |>
  transform(Sepal.Area = Sepal.Length * Sepal.Width)
```
The previous behaviour in this case was that only lines 2 and 3 would be sent, so you'd either have to go back to line 1 and re-send, or select all 3 lines before sending in visual mode. 

This also works in the case of mismatched parentheses. For example, hitting `<CR>` with the cursor on the second line here will now cause both lines to be sent:
``` r
(
123)
```

The new behaviour mimics RStudio, which feels more seamless to me.

I've tested in both R scripts and Quarto. 

I haven't tested with rnoweb. I understand R chunks in noweb are delimited at the start by lines like `<<my-chunk>>`, so when looking for the start of a noweb chunk I've just tested for lines beginning with `<<`.

### DIFF
--- a/lua/r/cursor.lua
+++ b/lua/r/cursor.lua
@@ -70,7 +70,7 @@ M.move_next_line = function()
         if filetype == "rnoweb" and string.sub(curline, 1, 1) == "@" then
             require("r.rnw").next_chunk()
             return
-        elseif (filetype == "rmd" or filetype == "quarto") and curline:find("^```$") then
+        elseif (filetype == "rmd" or filetype == "quarto") and curline:find("^```$") ~= nil then
             require("r.rmd").next_chunk()
             return
         end
--- a/lua/r/send.lua
+++ b/lua/r/send.lua
@@ -92,9 +92,7 @@ local function get_code_to_send(txt, row)
         ["program"] = true,
         ["brace_list"] = true,
     }
-    local is_root = function(n)
-        return root_nodes[n:type()] == true
-    end
+    local is_root = function(n) return root_nodes[n:type()] == true end
 
     while true do
         local parent = node:parent()
--- a/lua/r/send.lua
+++ b/lua/r/send.lua
@@ -36,6 +36,11 @@ local paren_diff = function(str)
     return llen1 - llen2
 end
 
+--- Check if line is a comment
+---@param line string
+---@return boolean
+local is_comment = function(line) return line:find("^%s*#") ~= nil end
+
 local M = {}
 
 --- Change the pointer to the function used to send commands to R.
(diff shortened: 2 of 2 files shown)


jalvesaq:
Thank you for your contribution! In a quick test, it seems to work and not introduce any bugs, including in an Rnoweb file. And I agree that it's more useful to send the whole chain of piped lines than only from the current line to the end. @PMassicotte has implemented a new way of sending part of a piped chain and might want to review this pull request too.

wurli:
Thanks for the review! 

I've just noticed one more potential improvement. If there's a comment in the middle of a pipeline like this, it stops all the lines getting sent. E.g. if you hit `<CR> on line 1 here, only lines 1 and 2 will get sent, and similar if you send from line 3:

``` R
iris |>
  subset(Species == "setosa") |>
  # Bla
  transform(Sepal.Area = Sepal.Length * Sepal.Width)
```

I think this should be quite easy to fix - are you happy for me to add another commit or would you prefer a separate pull request?

wurli:
Thank you for your review! I've just snuck in one more change to include commented lines. Now, all the lines in the following code would be sent if the cursor was placed on any of them:

``` R
# Bla
# Bla
iris |>
  subset(Species == "setosa") |>
  # Bla
  # Bla
  transform(Sepal.Area = Sepal.Length * Sepal.Width)
```

If you prefer I'd be more than happy to move this commit to its own pull request.

jalvesaq:
I think it's better to keep the new commits here. Thanks!

wurli:
Great, thanks! One more change - I've expanded the list of infix operators to cover the full list used by R, as I realised a few were missing. So now the following bits of code will all be sent in their entirety, regardless of where the cursor is:

``` R
2 %%
  4

4 %/%
  3

utils::
  head

1:
  10

!
  TRUE

a?
  b

list(a = 1)$
  a

list(a = 1)[
  "a"
]

list(a = 1)[[
  "a"
]]

# setClass("foo", representation(bar = "character"))
new("foo", bar = "baz")@
  bar
```

Think this will be it for now 😁

wurli:
Great, thanks! A couple more changes in line with RStudio's behaviour:
* I've expanded the list of infix operators to cover the full list used by R after realising a few were missing
* I've added some awareness of blank lines 

Now the following bits of code will all be sent in their entirety, regardless of where the cursor is:

``` R
2 %%

  4
```
``` R
1:

  10
```
``` R
!

  TRUE
```
``` R
a?

  b
```
``` R
list(a = 1)$

  a
```
``` R
list(a = 1)[

  "a"

]
```
``` R
# setClass("foo", representation(bar = "character"))
new("foo", bar = "baz")@

  bar
```

Think this will be it for now 😁

wurli:
I've temporarily closed this - I've noticed a couple more bugs I'd like to fix, then will reopen.

PMassicotte:
Let me know when it is ready to review/test!

wurli:
Hi both, thanks for bearing with. After some head scratching looking for an algorithm to reliably detect code to be sent, I decided to change tack and try using treesitter. This turned out to work really well!

Now:
* If the cursor is on a blank line, code is sent as if the cursor is on the next non-blank line
* If the cursor is on code, e.g. `1:10 |> sum()` or sum(list(a = 1)$a), the whole expression
  gets sent, including any interspersed linebreaks or comments
* Anything outside the current expression won't be sent:
    ``` R
    {              # With the cursor here, all 6 lines get sent
      1 + 1        # With the cursor here, only this line gets sent
      list(a = 1)$ # With the cursor here, lines 3 and 4 get sent
        a          # With the cursor here, lines 3 and 4 get sent
                   # With the cursor here, all 6 lines get sent
    }              # With the cursor here, all 6 lines get sent
    ```

The only slightly tricky bit was getting this to work with Quarto/R Markdown/Rnoweb, which I was able to get working by manually telling treesitter to parse the current chunk as R code. Based on my testing it all seems to work okay.

PMassicotte:
Thank you!

Rapidly browsed the code and it looks good. I will take a deeper look when I am back to home in 1-2 hr!

wurli:
Excellent, many thanks 🙂

PMassicotte (approved):
Looks good to me!

PMassicotte:
Tx for your contribution.

jalvesaq:
I'm seeing some warnings coming from the Lua server. The problems highlighted are two:

1.  `vim.treesitter.get_parser():parse()` receives only one argument (a table, according to treesitter's documentation). So, line 55 of `lua/r/send.lua` should be:

```lua
    vim.treesitter.get_parser(0, "r"):parse({ chunk_start_row, chunk_end_row })
```

2. `node` can be `nil` in lines 98, 103, and 105.

PMassicotte:
I will make a PR

wurli:
Thanks both! For future reference - how does one check the Lua server for warnings?

jalvesaq:
As far as I remember, when you run the Lua Language Server (`lua_ls`), the diagnostics are displayed by default, but diagnostics and completions of Neovim functions are available only because I have installed [neodev](https://github.com/folke/neodev.nvim). I also get additional diagnostics from `selene` enabled by [none-ls.nvim](https://github.com/nvimtools/none-ls.nvim). My complete configuration is in this [comment](https://github.com/R-nvim/cmp-r/issues/5#issuecomment-2000450625).

PR #104, opened 2024-03-31
Better hadling poissible nil values when using treesitter to send R code to the console (#99)

PMassicotte:
🐛 (send.lua): Fix bug where vim.treesitter.get_parser was not receiving an array as argument

♻️ (send.lua): Refactor get_code_to_send function to add nil checks for node variable to prevent potential null reference errors

See #99 for reference.


jalvesaq:
Thanks! The warnings are fixed, and it works.

PR #105, opened 2024-03-31
Support namespace prefix in `nvim.print()`

hongyuanjia:
### Pull request overview

Currently, `nvimcom::nvim.print()` does not support functions with namespace, e.g.:

```r
dplyr::filter(mtcars, mpg > 20)
           ^--- cursor here
 
> nvim.print("dplyr::filter", "mtcars")
Error in get(object) : object 'dplyr::filter' not found
In addition: Warning message:
In nvim.print("dplyr::filter", "mtcars") : object 'dplyr::filter' not found
```

This PR fixes this, besides other enhancements, including:

- Correct the logic for searching S3 Methods, i.e. using `cls` in the for loop, instead of `objclass`
- Only detect `.newobj` in the enclosing environment

### DIFF
--- a/nvimcom/DESCRIPTION
+++ b/nvimcom/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: nvimcom
-Version: 0.9.39
-Date: 2024-03-21
+Version: 0.9.40
+Date: 2024-04-01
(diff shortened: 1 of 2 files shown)


jalvesaq:
Thank you! Looks good for me! I tried it and it works.

PR #110, opened 2024-04-05
New feature: split file paths into its components

PMassicotte:
This PR adds two new keybindings:

1. `<localleader>sp`: Split a file path and wrap it around the `paste()` function.
2.  `<localleader>sh`: Split a file path and wrap it around the `here()` function.

This allows to breaks long lines to nicely format code.

Peek 2024-04-05 08-47

After, one can use the preferred formatter to wrap lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,18 @@ It's important to note that both functions require
 the [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter)
 plugin to be installed.
 
+For example, if you have the following file path under the cursor:
+
+```r
+read.csv("/home/user/Documents/file.csv")
+```
+
+Running the <localleader>sp command will transform it into:
+
+```r
+read.csv(paste("/home", "user", "Documents", "file.csv", sep = "/"))
+```
+
 ## Screenshots and videos
 
 None yet. Tell us if you published a video presenting R.nvim features.
--- a/README.md
+++ b/README.md
@@ -82,7 +82,6 @@ instructions on usage. See also the output of `:RMapsDesc`.
 
 ## Transitioning from Nvim-R
 
-
 During the conversion of VimScript to Lua, we decided to end support for features
 that were useful in the past but no longer sufficiently valuable to be worth
 the effort of conversion. We removed support for `Rrst` (it seems that not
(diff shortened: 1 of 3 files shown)


wurli:
😮 I like this! Worth considering other features in this area, e.g. [{codegrip}](https://github.com/lionel-/codegrip)-style reshaping?

One comment, should Windows path separators `\\` also be used to split paths? Both `/` and `\\` as file separators are valid if using R on Windows so this is pretty rare, but does happen.

PMassicotte:
> 😮 I like this! Worth considering other features in this area, e.g. [{codegrip}](https://github.com/lionel-/codegrip)-style reshaping?
> 
> One comment, should Windows path separators `\\` also be used to split paths? Both `/` and `\\` as file separators are valid if using R on Windows so this is pretty rare, but does happen.

I will check for windows support, I really hate seeing `\\` :) But you are right. As for _node_ formatting, I use  ["nvim-treesitter/nvim-treesitter"  (](https://github.com/Wansmer/treesj) which works very nicely. Could be an option here, but would introduce 1 dependency.

jalvesaq:
Good job, @PMassicotte! I see no problem in depending on `treejs`, but its README says that it already has a preset for R. Did you improve the preset?

PMassicotte:
`treejs` does not split path, it _simply_ reflows the code over multiple lines. At the moment, this PR do not need `treejs`, it will on only split the path. into components.

jalvesaq:
Looking at the split path:

- "Edit" would be a better category for the split commands because in the "Command" category are the commands sent to R Console.
- The command `require('r.path').split_path()` repeats the string "path". So, maybe the function name could be `split` instead of `split_path`.
- You use treesitter... People who don't have a tree-sitter parser for R will miss interesting `R.nvim` features. Should we make the availability of a tree-sitter parser for R a required dependency?
- A short description of the split path commands could be added to the list of new features in the _Transitioning from Nvim-R_ section in the README.

PMassicotte:
> * You use treesitter... People who don't have a tree-sitter parser for R will miss interesting `R.nvim` features. Should we make the availability of a tree-sitter parser for R a required dependency?

I think that would be nice. We have few functionalities that now relies on it.

New commit coming for your proposed changes.

jalvesaq:
Is it ready to be merged now?

PMassicotte:
Yes all good I think!

jalvesaq:
Thank you!

Issue #117, opened 2024-04-11
Open help documentation in a vertical split


cswingle:
Feature request: I would like help documentation to open in a vertical split, which is something that nvim-r used to do. There are hints that "vertical" is a legal option for the `nvimpager` option (lua/r/edit.lua, line 259), but it's not implemented in doc.lua.

I'll see if I can get it started in an upcoming pull request.

jalvesaq:
I forgot to delete the "vertical" option from the function that opens the example. But we can add the option back, although not in a complicated way as before.

PR #118, opened 2024-04-11
add nvimpager=vertical option, split help doc vertically. See #117

cswingle:
Adds "vertical" option to `nvimpager`, which causes help documentation to open in a vertical split instead of the default horizontal split.

Very lightly tested...

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,8 +117,8 @@ arguments that will be passed to `base::source()` and include the argument
 directly to the R Console without saving the code in a temporary file to be
 sourced (`max_lines_to_paste`).
 
-We reduced the number of options on how to display R documentation to: `"split"`,
-`"tab"`, `"float"` (not implemented yet), and `"no"`.
+The options for displaying R documentation (`nvimpager`) are now: `"split_h"`,
+`"split_v"`, `"tab"`, `"float"` (not implemented yet), and `"no"`.
 
 The options `openpdf` and `openhtml` were renamed as `open_pdf` and
 `open_html`, they now are strings and there with a minor change in how they
@@ -232,6 +232,6 @@ but temporary files are used in a few cases.
 - [colorout](https://github.com/jalvesaq/colorout): a package to colorize R's output.
 
 [cmp-r]: https://github.com/R-nvim/cmp-r
-[Neovim]: https://github.com/neovim/neovim
+[neovim]: https://github.com/neovim/neovim
 [southernlights]: https://github.com/jalvesaq/southernlights
 [colorout]: https://github.com/jalvesaq/colorout
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -1043,6 +1043,7 @@ The valid values of `nvimpager` are:
    "split"     : Split the window horizontally, as `:help` does. That is,
                  display R documentation above the editor window or at the
                  very top of the screen.
+   "vertical"  : Split the window vertically.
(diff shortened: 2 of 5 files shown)


wurli:
I'd suggest `split_h` and `split_v` instead of `split` and `vertical`? Would mean making superseding `split` in favour of `split_h`, and maybe deprecating it after a while.

jalvesaq:
You could use `elseif` to avoid increasing the indentation of the code.

jalvesaq:
We can delete this:

```lua
    if
        not config.nvimpager:find("tab")
        and not config.nvimpager:find("split")
        and not config.nvimpager:find("float")
        and not config.nvimpager:find("no")
    then
        warn(
            'Invalid `nvimpager` value: "'
                .. config.nvimpager
                .. '". Valid values are: "tab", "split", "float", and "no".'
        )
        return
    end
```

because `lua/r/config.lua` already warns if there is an invalid option.

jalvesaq:
> I'd suggest `split_h` and `split_v` instead of `split` and `vertical`? Would mean making superseding `split` in favour of `split_h`, and maybe deprecating it after a while.

I agree, but since we are still in the first months of `R.nvim` and most people still seem to be using `Nvim-R`, we can simply change the option name without a deprecation period. Anyway, we also have to document the change in the _Transitioning from Nvim-R_ section in the README.

jalvesaq:
The function `open_example()` (`lua/r/edit.lua`) has to be adjusted too. Ideally, we would have a single function to open examples and open the documentation, but the user might want to see and edit an example while looking at R documentation, and because the example can be edited, we can't simply replace it with another example or documentation. They need to be in separate functions, but we have to replicate the same algorithm in both.

cswingle:
Thanks for the comments. I'll revise a bit later today to change the option names, clean up the code, and add a short section to *Transitioning* explaining the options. In the section that actually does the `vsplit` I'm not doing any sanity checking of the terminal width/height, etc. I wasn't sure what the logic is in the horizontal split section so I didn't add anything similar to the new code.

jalvesaq:
I think it's better to avoid any complicated algorithm as we have in `Nvim-R`. If the user chooses "vertical" he/she will get a vertical split. The goal of the horizontal split is to behave exactly as Vim/Neovim default help, as described by:

```vim
:help help
```

cswingle:
I updated the options, removed the duplicate options checking, and added the list of current options in the *Transitioning* section of the README. I didn't modify the `split_h` code, which seemed to be working for me. `split_v` also works for me. I didn't test the section that opens a window for examples, but I did modify the code there. Let me know if anything else needs fixing/updating.

jalvesaq:
After `<LocalLeader>re` over a function name, I get this error:

```
Vim(lua):E5108:
Error executing lua /home/aquino/src/R.nvim/lua/r/edit.lua:261: attempt to compare number with nil
```

The problem is this code:

```lua
        if config.nvimpager == "split_v" then
            local wwidth = vim.fn.winwidth(0)
            local min_e = (config.editor_w > 78) and config.editor_w or 78
            local min_h = (config.help_w > 78) and config.help_w or 78
            if wwidth < (min_e + min_h) then nvimpager = "split_h" end
        end
```

which use config variables that no longer exist. I fixed it.

cswingle:
Thanks for fixing. I hadn't tested the examples function so I'm glad you did. I'll go read the manual to see what it does!

jalvesaq:

PR #126, opened 2024-04-26
add option `start_libs`

liubianshi:
When the option `start_libs` is set, it works as expected, but a warning pops up, "Unrecognized option \`start_libs\`."


jalvesaq:
Thank you for fixing my mistake!

PR #141, opened 2024-05-21
Fix RSendMotion key mapping

smb5:
Fix issues preventing the mapping for RSendMotion (\<LocalLeader\>m) from working as intended:

1. Remove the parentheses in require, as required by v:lua.
2. Don't append \<CR\> to the mapping as it will be used as the motion. Add a boolean parameter to create_maps to accomplish this.

### DIFF
--- a/lua/r/maps.lua
+++ b/lua/r/maps.lua
@@ -88,7 +88,7 @@ local map_desc = {
 ---@param plug string The "<Plug>" name.
 ---@param combo string Key combination.
 ---@param target string The command or function to be called.
----@param nocr boolean Don't append "<CR>" to target.
+---@param nocr boolean? Don't append "<CR>" to target.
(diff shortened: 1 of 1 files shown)


jalvesaq:
Thank you for your contribution! This is a feature that I don't use, but, in a quick test, the bug seems to be fixed.

I'm getting dozens of warnings from the Lua language server (or Selene) about the function `create_maps` requiring five arguments. The warnings disappear if a question mark is added to the fifth argument's documentation:

```lua
---@param nocr boolean? Don't append "<CR>" to target.
```

Could you change `boolean` to `boolean?`, please?

jalvesaq:
I added the quotation mark. Thanks for your contribution!

smb5:
Thanks! Beat me to it.

PR #144, opened 2024-05-26
Add vim.ui.open support for pdf opening

Ben10164:
In Neovim 0.10, a new function `vim.ui.open` was added which opens ANY file with the system's configured application for said file's filetype. 

As I'm not very familiar with the structure of this project, this PR just adds a line and suppresses an error. I implemented it quickly by letting the user specify `pdfviewer = ""` in their config. My hope is that someone with more familiarity to the layout and style of this repo could help me implement this in a way consistent with the repo.

(I understand that using `vim.ui.open` will most likely result in a loss of support for SyncTeX, but the point of this PR is to start discussing basic opening ability with the new command)

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -1482,14 +1482,12 @@ Wayland windows managers other than Sway.
 
 R.nvim will call Sumatra to open the PDF on Windows, Skim on Mac OS X, and
 Zathura on Linux, but you can change the values of `synctex` and `pdfviewer`
-in your config if you don't need SyncTeX support. For systems using 
-Neovim v10.0+, setting `pdfviewer` to an empty string will open the PDF with 
-the OS default application. Example:
+in your config if you don't need SyncTeX support. Setting `pdfviewer` to an
+empty string will open the PDF with the OS default application. Examples:
 >lua
    synctex = false
-   pdfviewer = "open"     -- Default PDF viewer (macOS or Windows)
-   pdfviewer = "xdg-open" -- Default PDF viewer (Linux)
-   pdfviewer = ""         -- Default PDF viewer (Any) [Requires Neovim v0.10+]
+   pdfviewer = "evince"   -- Use Evince as PDF viewer
+   pdfviewer = ""         -- Default PDF viewer (and no SyncTeX support)
 <
 If editing an Rmd file, you can produce the html result with <LocalLeader>kr
 or <LocalLeader>kh (it may also work with other file types). The html file
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -1482,7 +1482,9 @@ Wayland windows managers other than Sway.
 
 R.nvim will call Sumatra to open the PDF on Windows, Skim on Mac OS X, and
 Zathura on Linux, but you can change the values of `synctex` and `pdfviewer`
-in your config if you don't need SyncTeX support. Example:
+in your config if you don't need SyncTeX support. For systems using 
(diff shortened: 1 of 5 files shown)


jalvesaq:
The current Ubuntu Linux distribution ships Neovim 0.9.5. So, I think we can implement features that depend on Neovim 0.10, but it's better to keep compatibility with 0.9.5.

I think it makes sense to use `""` to open the default PDF viewer, but users should be able to find this information in the documentation. So, I think your pull request needs to:

- check if Neovim version is equal or higher than `0.10` (`vim.fn.has("nvim-0.10")`).
- document in `doc/R.nvim.txt` that if Neovim `>=` 0.10, `""` can be used as an alternative to `open` or `xdg-open`.

In the future, we can remove reference to `open`/`xdg-open` from the documentation.

Ben10164:
> * check if Neovim version is equal or higher than `0.10` (`vim.fn.has("nvim-0.10")`).

Added in latest commit!

> * document in `doc/R.nvim.txt` that if Neovim `>=` 0.10, `""` can be used as an alternative to `open` or `xdg-open`.

Also added in latest commit, but let me know if you want me to go into further detail. Currently it is: 

```txt
R.nvim will call Sumatra to open the PDF on Windows, Skim on Mac OS X, and
Zathura on Linux, but you can change the values of `synctex` and `pdfviewer`
in your config if you don't need SyncTeX support. For systems using 
Neovim v10.0+, setting `pdfviewer` to an empty string will open the PDF with 
the OS default application. Example:
>lua
   synctex = false
   pdfviewer = "open"     -- Default PDF viewer (macOS or Windows)
   pdfviewer = "xdg-open" -- Default PDF viewer (Linux)
   pdfviewer = ""         -- Default PDF viewer (Any) [Requires Neovim v0.10+]
<
```
(Starts at line `1483`)

jalvesaq:
It's OK now! Thanks!

We only have cosmetic complains from `selene` and `stylua`. To silence them:

```lua
    if vim.fn.has("nvim-0.10") == 0 and config.pdfviewer == "" then
        warn(
            'R.nvim: Having a `pdfviewer` value of `""` is only supported for Neovim 0.10+.'
        )
    end

    if vim.fn.executable(config.pdfviewer) == 0 and config.pdfviewer ~= "" then
        warn(
            "R.nvim: Please, set the value of `pdfviewer`. The application `"
                .. config.pdfviewer
                .. "` was not found."
        )
    end
```

and

```lua
    if config.pdfviewer == "" then
        vim.ui.open(fullpath)
        return
    end
```

jalvesaq:
It's OK now! Thanks!

We only have cosmetic complains from `selene` and `stylua`. To silence them:

```lua
    if vim.fn.has("nvim-0.10") == 0 and config.pdfviewer == "" then
        warn(
            'R.nvim: Having a `pdfviewer` value of `""` is only supported for Neovim 0.10+.'
        )
    end

    if vim.fn.executable(config.pdfviewer) == 0 and config.pdfviewer ~= "" then
        warn(
            "R.nvim: Please, set the value of `pdfviewer`. The application `"
                .. config.pdfviewer
                .. "` was not found."
        )
    end
```

and

```lua
    if config.pdfviewer == "" then
        vim.ui.open(fullpath)
        return
    end
```

folke:
You can just do:

```lua
      require("r.pdf.generic").open = vim.ui.open
```

Ben10164:
@jalvesaq 
Can confirms it works on MacOS!

PR #157, opened 2024-06-25
Add a new module to replace all the '$' subsetting operators with '[[' in the current buffer

PMassicotte:
This PR add a new module to replace all the `$` subsetting operators with `[[` in the current buffer.  The new keybinding is `<localleader>cs` (change subsetting).

This works with any number of nested subset:

```r
x$a

x$a$b$c$d
```

Will become:

```r
x[["a"]]

x[["a"]][["b"]][["c"]][["d"]]
```

I also created a new folder named `format` that includes all the new lua modules dedicated to format R code to comply with the `lintr` package.


jalvesaq:
Great job! The only thing missing is the documentation.

PMassicotte:
oups! will do that!

PMassicotte:
To make sure I do things correctly for the documentation. Should I modify directly the `R.nvim.txt` by adding a new section?

jalvesaq:
> To make sure I do things correctly for the documentation. Should I modify directly the R.nvim.txt by adding a new section?

Yes, we have no other similar feature. So, its documentation would be in a new section.

jalvesaq:
Opss. I thought that this was about installing packages...

PMassicotte:
Was asking because I was wondering if this file was automatically generated based on function headers (likr using roxygen).

jalvesaq:
I think that section 6.36. can have a general title to accommodate both "Converting numbers to explicit integers" and "Replace subsetting operators".

PMassicotte:
I modified the help section. Let me know if you think it needs more improvement.

jalvesaq:
It's good! Thanks!

PR #160, opened 2024-06-27
Automatically install missing packages in the current buffer

PMassicotte:
This PR uses `lintr` to automatically install missing packages used in the current buffer. It will detect any missing packages used either as:

```r
library(package)
```
or as:

```r
package::fun()
```


jalvesaq:
Great job again!

Issue #147, opened 2024-06-14
Can not get arguments completion use `devtools::load_all` of functions which are not export in NAMESPACE


shaman-yellow:
This may not be a issue, but is a different feature from Nvim-R.
`devtools::load_all` has parameter of `export_all`, theoretically the functions were exported, but I can't get argument completion unless I specify in `NAMESPACE` that the function is exported.
For me, since my package may not be shared with others, I will not specifically indicate in NAMESPACE that the function needs to be exported. `devtools::load_all` makes all these functions exported. This is very convenient.

jalvesaq:
The completion data includes only functions from libraries listed by the `search()` command: https://github.com/R-nvim/R.nvim/blob/dc4eb50e4fd11b342d89b50b66b7dff4166f3673/nvimcom/src/nvimcom.c#L816

I don't use `devtools`. So, you or someone else would have to do a pull request adding the ability to list functions loaded by `devtools::load_all()` and other people from the `R-nvim` organization should approve the pull request.

shaman-yellow:
But `Nvim-R` does. What's the difference?
I did not notice a significant difference between <R.nvim/nvimcom/src/nvimcom.c>
with <Nvim-R/nvimcom/src/nvimcom.c>

jalvesaq:
I don't know. Could you describe simple steps that I could copy and paste to see what happens? Anyway, I will not have free time in the next few days.

jalvesaq:
Just a random idea: if you have installed your package in the past as a normal library, Nvim-R might have built the necessary cache files at `~/.cache/Nvim-R`...

shaman-yellow:
We can use `usethis::create_package` to create a test package, then load it by `devtools::load_all`:

```R
usethis::create_package("~/ATestPackage")
writeLines("ATestFun <- function(x) {\n print('Test this') \n}", "~/ATestPackage/R/test.R")
list.files("~/ATestPackage/", recursive = T)
# [1] "DESCRIPTION" "NAMESPACE"   "R/test.R"   

# The `NAMESPACE` is empty
readLines("~/ATestPackage/NAMESPACE")
# [1] "# Generated by roxygen2: do not edit by hand"
# [2] ""                                            

# Now we load it
devtools::load_all("~/ATestPackage")

# The `ATestPackage` exists in `search()` results
search()[grep("ATestPackage", search())]
# [1] "package:ATestPackage"

# we can directly call it without prefix `ATestPackage::`
ATestFun()
[...]

PR #165, opened 2024-06-29
Restore debugging feature

jrwishart:
Allow line highlighting in source buffer during browser and debug calls
of functions.


jrwishart:
Apologies for the delay. Full time work was prioritised until this weekend. I removed some code paths that looked like it was unsupported tmux.

PMassicotte:
Thank you for this PR!

I am testing it on my end, and while it appears to enter debug mode, I do not see any visual indication (e.g., line highlighting) of the debug process.

Peek 2024-06-29 06-45

jrwishart:
Thanks for the quick feedback.

Admittedly, I have never tried to debug a function sourced inline into the global environment. I have been doing it solely from a loaded package for package development. e.g. via `devtools::load_all()`. 

It works fine when functions are loaded (see video).

https://github.com/R-nvim/R.nvim/assets/4605715/dac96680-d4ea-4df7-883c-dc2f670a8cb1

I can reproduce the same behaviour you have that the functions sourced inline in the console will not trigger the visual highlighting debug/browser.

jrwishart:
Might be related to https://github.com/R-nvim/R.nvim/issues/147. Perhaps `nvimcom` doesn't search functions that are sourced into the global env for the debug lines there?

jrwishart:
Just saw the CI checks failing. I'll see if I can fix those

doc/R.nvim.txt
```diff
@@ -549,6 +550,78 @@ put in your `init.lua`:
...
+<
+One limitation is that R-Nvim also does not help you setting breakpoints.
+You have to set them manually as explained above.
+
+Another limitation of R-Nvim is that it might be unable to jump to the line in
+the source code that is being evaluated if either you have sent the function
+directly to the R Console or the function has been sent by the `knitr`
+package, without calling the function `source()`. In either case, R-Nvim will
+try to find the pattern `funcname <- function(` in the buffer being edited. If
+the pattern is found, the cursor will jump to its position.
```
jrwishart, line 618: Seems like it was already a known implementation? (I didn't write the documentation in this PR, I am merely restoring it from the previous state).

jrwishart: I just tried sourcing the function in a file using the `source` function and I get the same behaviour (no highlighting). So perhaps the documentation could be updated to reflect that loading functions into the environment via `devtools` will work but not using `source`?

PMassicotte:
Peek 2024-06-29 08-22

doc/R.nvim.txt
PMassicotte: If I understand correctly, debugging will only work when sourcing a package with `devtools`.

PMassicotte:
I am trying with a package, and still can not get it working. Any ideas?

Peek 2024-06-29 09-15

PMassicotte:
Maybe because I am using an external terminal?
f

jalvesaq:
Some notes:

- I have never used `devtools`. I only install it when someone reports a bug that requires it to be installed.
- When using the debugging feature, I manually sourced the file with the function I wanted to debug. This way, R knew where the function was from.
- The terminal emulator should not affect the process because `nvimcom` reads the R prompt (which indicates the debugging mode) directly from R's memory.

lua/r/debug.lua
```diff
@@ -0,0 +1,147 @@
...
+    s.func_offset = -1 -- Not found
+    local sbopt = vim.o.switchbuf
+    vim.o.switchbuf = "useopen,usetab"
+    local curtab = vim.fn.tabpagenr()
+    local isnormal = vim.fn.mode() == "n"
+    local curwin = vim.fn.winnr()
+    r_bufnr = vim.api.nvim_get_current_buf()
+    vim.cmd("sb " .. r_bufnr)
+    vim.fn.sleep(30) -- Time to fill the buffer lines
+    rlines = vim.fn.getline(1, "$")
```
jrwishart, line 47: In this PR, the debug code only finds `rlines` (The R source lines I believe) if there is no `external_term` in the config. 

There was vimscript code found `rlines` when debugging and I could look to convert it. That code is here. 

https://github.com/R-nvim/R.nvim/blob/5b551860efe6a07b2e565aa24e124612e18c0a5e/R/start_r.vim#L97-L102

I searched the commit history last week but couldn't determine what the proper way to do this in the lua plugin. As such I removed it. I mentioned it in the initial comment that I removed some code paths that looked like unsupported tmux.

Be happy to be advised on how to fix.

jalvesaq:
I'm trying to fix the pull request...

jalvesaq:
It's fixed on my side if R is running in an external terminal emulator and the function being debugged is in another script which was read with the `source()` command. Here's how to try it:

Uninstall `nvimcom`:

```sh
echo 'remove.packages("nvimcom")' | R
```

Then, follow the instructions from the documentation:

For example, if you have a file named `my_function.R` with the following contents:

```r
add.one <- function(x) {
    y <- x + 1
    print(y)
    return(invisible(NULL))
}
```

You could debug the `add.one()` function from another script such as:

```r
source("path/to/my_function.R")
debug(add.one)
add.one(7)
```

jalvesaq:
I removed "r" and "debug" from the names of Lua functions because the information that the function is related to debugging R functions is already present in the module name, "r.debug".

jrwishart:
Thanks for the quick update. I can confirm that it works for `source` now too 👍 

I took the liberty to update the documentation to show the `devtools` approach. The only loose end is the last commit you made seems to have broken the cursor staying in the R console. Previously I could stay in the R console while debugging but now it is staying in the highlighted source line on every step requiring manually switching cursor to the R console each time.


https://github.com/R-nvim/R.nvim/assets/4605715/a663f124-7d62-4304-a953-4f8d1187db2d

lua/r/debug.lua
jalvesaq: At this point, it's better to no longer look at `Nvim-R` code. What has to be done is relatively simple:

1. When beginning the debugging, load the file where the code is or jump to it if it's already open (this is already done in the specific case that I fixed) and store the buffer number (not done yet).
2. From the second jump onward, move the cursor without switching to its buffer (this will be different from `Nvim-R`).

jalvesaq:
>  The only loose end is the last commit you made seems to have broken the cursor staying in the R console.

I can look at this on the next weekend. I think the solution will be to implement steps 1 and 2 from my previous comment.

PMassicotte:
No luck on my side.
Peek 2024-06-30 06-44

jalvesaq:
Try an R script with these lines:

```r
source("debug.R")
debug(addone)
addone(1)
```

In its current buggy state, it only works if the function is in *another* script which still has to be open.

PMassicotte:
Same thing, nothing happens.
Peek 2024-06-30 08-16

jalvesaq:
It only works if R is running in an external terminal emulator or a Tmux split pane because the code to open and switch buffers is yet to be fixed.

jalvesaq:
@jrwishart, please, see https://github.com/R-nvim/R.nvim/discussions/166 as an alternative to this pull request...

jalvesaq:
I found some time today and fixed the bugs I knew existed. Of course, instead of merging this pull request, it would be better to have a real DAP plugin.

jrwishart:
The last commit fixed the cursor position but seems to have introduced a new bug where it doesn't find the correct line if debug or browser is called from another function in a different file. E.g. in the video, `bar` calls `hi`, `bar` is in `foo.R` and `hi` is in `qux.R` and it gets the wrong line match for `hi` if it is a nested browser/debug call.

https://github.com/R-nvim/R.nvim/assets/4605715/204a8ae3-8500-4664-9945-283a169ec5d9

jalvesaq:
I think the problem is in line 90:

```lua
    if not s.debugging then
```

The buffer and window numbers are set only when the debugging session begins. I can look at this at the end of the week, but, please, feel free to fix it. It will be better for the `R.nvim` project if I'm not the only one to be able to maintain the debugging code...

jalvesaq:
Now, it should jump to other functions if you step into them.

jalvesaq:
If this pull request were merged, I believe the debug functionality of `R.nvim` would be as good (but also as limited) as that of `Nvim-R`. We only have to wait a few days for a working DAP for R before deciding either to merge or to close this pull request.

PMassicotte:
I agree. I will try this PR tomorrow when I am back to the office.

jrwishart:
Thanks for fixing. I was planning on having a close look after my business trip finishes at the end of the week but you were too fast again 🥇  

I am happy to wait a little bit for the DAP situation to become clearer and avoid merging the PR here for now. The thread in https://github.com/dgkf/debugadapter looks promising. I'm willing to use R-devel if DAP works there. Can always pivot to using this PR if I need R release in the interim.

PMassicotte:
Ok I have some time to try it. How this is supposed to work? Can I simply add a `browser()` somewhere in a script and source it?

PMassicotte:
It works on my side if I do `srouce` + `debug(fun)` in the terminal and then call the function. Should it works also if I place a `browser()` somewhere in a script and then use the send all lines to the terminal?

jalvesaq:
Thanks for catching this! It should work now.

jalvesaq:
When we `source()` a script, R knows where a function comes from and this information is available for us in an internal variable that can be accessed from C. When we send the function directly to R Console, R doesn't know where the function comes from. Then, we have to get the contents of R Console and look for the string `"^debugging in: "` or `"Called from: "`, extract the function name, and seek for it in the R script currently being edited.

So, the debugging will not work if you send a function directly to the R Console and switch to another buffer before starting to debug the function.

jalvesaq:
There is still one possible improvement. When debugging a function from an installed package or a function sent directly to the R Console from another R script, the source code isn't available, but the whole function is printed in the R Console, and we could copy it from there and paste it in a new temporary R script. However, before investing time in doing this, let's see if the DAP works...

PMassicotte:
> When we `source()` a script, R knows where a function comes from and this information is available for us in an internal variable that can be accessed from C. When we send the function directly to R Console, R doesn't know where the function comes from. Then, we have to get the contents of R Console and look for the string `"^debugging in: "` or `"Called from: "`, extract the function name, and seek for it in the R script currently being edited.
> 
> So, the debugging will not work if you send a function directly to the R Console and switch to another buffer before starting to debug the function.


Great this is working if I create a script like this:

```r
addone <- function(x) {
  x + 1
}


browser()
addone(10)
```

and then `source()` it from the console.

wurli:
Hi all, I've only been somewhat following this thread because I very rarely use R's debugger. But just want to throw this out there in case it's relevant - the guys at Posit are working on a DAP server/LSP server for R called [Ark](https://github.com/posit-dev/ark). It's not yet usable by other software than their new Positron IDE, but the README says it will be soon.

jalvesaq:
Thanks for the news, @wurli! This is another reason for not merging this pull request, but we can keep it open until either `Ark` or `debugadapter` is ready to be used in Neovim.

akthe-at:
@wurli @jalvesaq It looks like the Ark project is moving along and is starting to be usable outside of positron (Jupyter)...Very exciting times

jalvesaq:
Ark is still usable only by Positron.

https://codeberg.org/dgkf/debugadapter still says "Currently, nothing will happen" when "debugging".

Do we have other options? Should we keep waiting or merge this pull request?

jalvesaq:
I'm resolving the conflicts...

jalvesaq:
I fixed the conflicts in Github's online editor.

lua/r/debug.lua
```diff
@@ -0,0 +1,199 @@
...
+        local run_cmd = {
+            "tmux",
+            "-L",
+            "Rnvim",
+            "capture-pane",
+            "-p",
+            "-t",
+            require("r.external_term").get_tmuxsname(),
+        }
+        local resp = require("r.utils").system(run_cmd, { text = true }):wait()
```
PMassicotte: This is the faulty line when trying to source a file that includes a `browser()` call.

jalvesaq:
Fixed: `utils.system()` replaced with `vim.system()`

doc/R.nvim.txt
```diff
@@ -566,6 +567,83 @@ put in your `init.lua`:
...
+   1. Load the functions into R either by a) sourcing the R script where the
+      function is with the `source()` function of the `base` package or b)
+      loading all desired functions with `devtools::load_all(<relevant path>)`.
+
+   2. Flag a function or line of code within a function to use the debugger.
+      This can be done by using the `debug` functions or setting a breakpoint
+      with the `browser` functions from the `base` package. To use `debug`,
+      in the R Console, run either `debug(funcname)` or `debugonce(funcname)`
+      where `funcname` is the name of the function (not quoted). R.nvim will
+      send this command to R Console if you press <LocalLeader>db in Normal
```
PMassicotte, line 586:
```suggestion
      send this command to R Console if you press <LocalLeader>bg in Normal
```

doc/R.nvim.txt
```diff
@@ -566,6 +567,83 @@ put in your `init.lua`:
...
+      mode with the cursor over the function name. An example of using
```
PMassicotte, line 587:
```suggestion
      mode with the cursor over the function name. Use `<LocalLeader>ud` with the cursor on a function to undebug it. An example of using
```

PMassicotte:
Thanks, fixed small error on keymaping and added a sentence.

PR #172, opened 2024-07-04
Fix new treesitter split path

PMassicotte:
This fixes the split path function due to changes in `treesitter-r`.


PMassicotte:
I messed with git. Will close it and clean it.

PR #174, opened 2024-07-04
Fix the node names to correctly capture file path

PMassicotte:
This is to adapt to the re-write of `treesitter-r`.

🐛 (path.lua): fix node type check from "string" to "string_content" to correctly identify string nodes
♻️ (path.lua): refactor to handle parent node extraction and path replacement more robustly


jalvesaq:
Thanks! I confirm that it fixes the code.

Issue #175, opened 2024-07-05
Has function of `<Leader>l` and `<Leader>d` been changed?


joihci:
When I press `<Leader>l` or `<Leader>d` inside a function body the whole function is sent to terminal instead of the line under cursor. When cursor is outside of a function these keybinds work as expected. However using the motion command inside a function works i.e. `<Leader>ml` sends the line under cursor, `<Leader>md` does nothing.

Is this a bug or has the function of `<Leader>l` and `<Leader>d` been changed?

See behaviour on provided screenshot:
![image](https://github.com/R-nvim/R.nvim/assets/105873203/c97f870c-7e39-4e80-bd1d-adbecc44fa9f)

Info:
```
NVIM v0.10.0
Build type: Release
LuaJIT 2.1.1713484068

OS: Ubuntu 22.04.4 LTS
```

Using minimal init.vim for provided example in screenshot:
```
call plug#begin(stdpath("data")."/plugged")
Plug 'R-nvim/R.nvim'
Plug 'nvim-treesitter/nvim-treesitter', { 'do': ':TSUpdate' }
call plug#end()

lua require("r").setup({ pipe_keymap = "<M-,>" })
let mapleader="\<Space>"
let maplocalleader="\<Space>"
lua <<EOF
require("nvim-treesitter.configs").setup {
  ensure_installed = "all", -- one of "all",
  highlight = {
    enable = true,
    additional_vim_regex_highlighting = false,
  },
  playground = {
    enable = true,
    disable = {},
    updatetime = 25,
    persist_queries = false,
    keybindings = {
      toggle_query_editor = "o",
      toggle_hl_groups = "i",
      toggle_injected_languages = "t",
      toggle_anonymous_nodes = "a",
      toggle_language_display = "I",
      focus_language = "f",
      unfocus_language = "F",
      update = "R",
      goto_node = "<cr>",
      show_help = "?",
    },
  }
}
EOF
```

PMassicotte:
Probably related to changes in `treesitter-r`.

joihci:
Yes, it's convenient to execute a line in a function body for quick debugging and besides there is a keybind for sending the current function under cursor (shouldn't matter whether cursor is in body or on definition: `<Leader>fc`.

But now when I press `<Leader>fc` within a function I get an error (had not tried it before creating this issue):
![image](https://github.com/R-nvim/R.nvim/assets/105873203/18e8085d-7aa0-4fcc-ab9b-7f2d3a125575)

[...]

PR #176, opened 2024-07-05
Fix send function treesitter r

PMassicotte:
Only sourcing function once (#175).

### DIFF
--- a/lua/r/send.lua
+++ b/lua/r/send.lua
@@ -662,9 +662,6 @@ M.funs = function(bufnr, capture_all, move_down)
         [[
     (binary_operator
       (function_definition)) @rfun
-
-    (binary_operator
-      (function_definition)) @rfun
     ]]
     )
 
--- a/lua/r/send.lua
+++ b/lua/r/send.lua
@@ -660,10 +660,10 @@ M.funs = function(bufnr, capture_all, move_down)
     local r_fun_query = vim.treesitter.query.parse(
         "r",
         [[
-    (left_assignment
+    (binary_operator
       (function_definition)) @rfun
 
-    (equals_assignment
+    (binary_operator
       (function_definition)) @rfun
     ]]
     )
--- a/lua/r/send.lua
+++ b/lua/r/send.lua
@@ -685,17 +685,12 @@ M.funs = function(bufnr, capture_all, move_down)
     for id, node in r_fun_query:iter_captures(root_node, bufnr, 0, -1) do
         local name = r_fun_query.captures[id]
 
-        -- Kinda hacky, but it works. Check if the parent of the function is
-        -- the root node, if so, it's a top level function
-        local s, _, _, _ = node:parent():range()
-
-        if name == "rfun" and s == 0 then
+        if name == "rfun" and node:parent() == root_node then
             local start_row, _, end_row, _ = node:range()
 
             if
                 capture_all or (cursor_pos >= start_row + 1 and cursor_pos <= end_row + 1)
             then
-                M.source_lines(lines, nil)
                 lines = vim.fn.extend(
                     lines,
                     vim.api.nvim_buf_get_lines(bufnr, start_row, end_row + 1, false)
@@ -707,6 +702,7 @@ M.funs = function(bufnr, capture_all, move_down)
             end
         end
     end
+
     if #lines then M.source_lines(lines) end
 end
 


jalvesaq:
On my side, with plugins updated and the cursor within a function, the command to send a line and go to the next line sends the whole function while the command to send the entire function sends an empty line.

jalvesaq:
The development of `nvim-treesitter` frequently introduces breaking changes. So, we need tests for our features that depend on `nvim-treesitter`. It would be better if these tests could be run automatically, but I'm not good with automatic tests. So, I'm thinking of something like the script below to manually test the algorithm to send lines. We say in comments where the cursor has to be placed and what is the expected result of `RDSendLine`:

```r
# Cursor on `d`. Send the next line.
d <- data.frame(var1 = c(1, 2, 3, 4, 5), var2 = c(1, 2, 3, 3, 2))

# Cursor on `if`. Send the next two lines.
if (TRUE)
    library("ggplot2")

# Cursor on `if`. Send the bracket block.
if (TRUE) {
    # Cursor on `print`. Send only the line or the entire bracket block?
    print(1)
    print(2)
}

# Cursor on `plus_one`. Send the bracket block.
plus_one <- function(a) {
    # Cursor on `b`. Send only the line or entire function?
    b <- a + 1
    b
}

# Cursor anywhere in the four lines below. Send the four lines.
ggplot(d, aes(var1, var2)) +
    geom_point() +
    xlab("X") +
    ylab("Y")
```

The questions in the comments mean that I'm not sure what should be the result. Perhaps, different results for `RDSendLine` and `RSendLine`? For example, when the cursor is within a bracket block, `RDSendLine` could send only the current line, and `RSendLine` could send the entire block. Or, perhaps, something else...

PMassicotte:
> On my side, with plugins updated and the cursor within a function, the command to send a line and go to the next line sends the whole function while the command to send the entire function sends an empty line.

`localleader<fc>` is not sending the whole function on your side? Seems to be ok on my side.

Here I am using `fc` and `fd`:

Peek 2024-07-06 06-26

Ere

PMassicotte:
> The development of `nvim-treesitter` frequently introduces breaking changes. So, we need tests for our features that depend on `nvim-treesitter`. It would be better if these tests could be run automatically, but I'm not good with automatic tests. So, I'm thinking of something like the script below to manually test the algorithm to send lines. We say in comments where the cursor has to be placed and what is the expected result of `RDSendLine`:
> 
> ```r
> # Cursor on `d`. Send the next line.
> d <- data.frame(var1 = c(1, 2, 3, 4, 5), var2 = c(1, 2, 3, 3, 2))
> 
> # Cursor on `if`. Send the next two lines.
> if (TRUE)
>     library("ggplot2")
> 
> # Cursor on `if`. Send the bracket block.
> if (TRUE) {
>     # Cursor on `print`. Send only the line or the entire bracket block?
>     print(1)
>     print(2)
> }
> 
> # Cursor on `plus_one`. Send the bracket block.
> plus_one <- function(a) {
>     # Cursor on `b`. Send only the line or entire function?
>     b <- a + 1
>     b
> }
> 
> # Cursor anywhere in the four lines below. Send the four lines.
> ggplot(d, aes(var1, var2)) +
>     geom_point() +
>     xlab("X") +
>     ylab("Y")
> ```
> 
> The questions in the comments mean that I'm not sure what should be the result. Perhaps, different results for `RDSendLine` and `RSendLine`? For example, when the cursor is within a bracket block, `RDSendLine` could send only the current line, and `RSendLine` could send the entire block. Or, perhaps, something else...

Some ideas around here: https://github.com/Wansmer/treesj/blob/main/tests/langs/r/r_split_spec.lua

jalvesaq:
I confirm that `\fc` works if the cursor is on an _empty_ line, as in your GIF.

> Some ideas around here: https://github.com/Wansmer/treesj/blob/main/tests/langs/r/r_split_spec.lua

Yes, something like this or the tests that @she3o were developing.

PMassicotte:
> I confirm that \fc works if the cursor is on an empty line, as in your GIF.

It also works when a non-empty line. Is it not working for you?

Peek 2024-07-06 09-31

jalvesaq:
It's working now. Probably there was something wrong on my side. I'm sorry for the noise.

jalvesaq:
I have to make the changes below to make the pull request work:

```diff
diff --git a/lua/r/send.lua b/lua/r/send.lua
index 223f750..35db34e 100644
--- a/lua/r/send.lua
+++ b/lua/r/send.lua
@@ -742,18 +742,11 @@ M.funs = function(bufnr, capture_all, move_down)
 
     for id, node in r_fun_query:iter_captures(root_node, bufnr, 0, -1) do
         local name = r_fun_query.captures[id]
-
-        -- Kinda hacky, but it works. Check if the parent of the function is
-        -- the root node, if so, it's a top level function
-        local s, _, _, _ = node:parent():range()
-
-        if name == "rfun" and s == 0 then
+        if name == "rfun" then
             local start_row, _, end_row, _ = node:range()
-
             if
                 capture_all or (cursor_pos >= start_row + 1 and cursor_pos <= end_row + 1)
             then
-                M.source_lines(lines, nil)
                 lines = vim.fn.extend(
                     lines,
                     vim.api.nvim_buf_get_lines(bufnr, start_row, end_row + 1, false)
```

- Don't check `s == 0` because the function might not begin on the first line of the buffer.
- Don't source the lines within the loop, only after it (avoid sending the functions more than once).

jalvesaq:
There is a problem in my "fix". Inner functions will be sent separately. But `s` wasn't `0` if the function doesn't begin on line `1`. Maybe the node is not the root yet.

jalvesaq:
This seems to fix it:

```diff
diff --git a/lua/r/send.lua b/lua/r/send.lua
index 223f750..369e18d 100644
--- a/lua/r/send.lua
+++ b/lua/r/send.lua
@@ -742,18 +742,11 @@ M.funs = function(bufnr, capture_all, move_down)
 
     for id, node in r_fun_query:iter_captures(root_node, bufnr, 0, -1) do
         local name = r_fun_query.captures[id]
-
-        -- Kinda hacky, but it works. Check if the parent of the function is
-        -- the root node, if so, it's a top level function
-        local s, _, _, _ = node:parent():range()
-
-        if name == "rfun" and s == 0 then
+        if name == "rfun" and node:parent() == root_node then
             local start_row, _, end_row, _ = node:range()
-
             if
                 capture_all or (cursor_pos >= start_row + 1 and cursor_pos <= end_row + 1)
             then
-                M.source_lines(lines, nil)
                 lines = vim.fn.extend(
                     lines,
                     vim.api.nvim_buf_get_lines(bufnr, start_row, end_row + 1, false)
```

PMassicotte:
Great! Can you send the changes to the PR directly?

jalvesaq:
Thanks!

PR #179, opened 2024-07-07
Fix some auto-completion bugs

### LINKED ISSUE #123: Open/close parenthesis no longer works when using autocomplete?
Is it just me, but it seems that the opening/closing parenthesis are not working when selecting a function from the completion menu?

I think before it was inserting `()` and placing the cursor right inside ready to use arguments. It seems that I have to type them manually now.

Peek 2024-04-19 13-45

jalvesaq:
- Show the default values of arguments.
- Show the summary of `data.frame` columns of the `datasets` package as well as of elements of other packages' lists.
- Close #123

### DIFF
--- a/nvimcom/DESCRIPTION
+++ b/nvimcom/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: nvimcom
-Version: 0.9.45
-Date: 2024-07-06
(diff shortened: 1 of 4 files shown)


PMassicotte:
Looks good to me!

PR #187, opened 2024-07-15
Use treesitter to get code language

jalvesaq:
This pull request replaces the `b:IsInRCode` functions with a single tree-sitter based function. It seems to work well with Rmarkdown or Quarto, but not with Rnoweb. We still need a specific `is_in_R_code` function for the `rhelp` file type because there is no tree-sitter grammar for it.


jalvesaq:
I fixed `get_lang()` by not using tree-sitter for `rnoweb` and `rhelp`. I also removed the `vim.notify()`.

jalvesaq:
This pull request is ready for review now. It:

- Avoids the complication of Lua functions called via buffer variables (`b:IsInRCode`) which was OK in VimScript but unusual in Lua. We still have two functions specific to `rnoweb` and `rhelp` in `utils.lua`. The `rnoweb` function can be deleted if either the `rnoweb` parser is improved or we find the correct way of getting the language at the cursor position. The `rhelp` function can be deleted if someone creates a parser for the language.
- Fixes sending code from `rhelp` files to R. There is an additional function specific to `rhelp` with two auxiliary functions in `send.lua`. The three functions can be deleted if a parser for `rhelp` is created.

jalvesaq:
Note: the tests fail to run correctly (`make test`) and I don't know how to fix them.

PMassicotte:
Excellent, it simplifies the code a lot. Testing/reviewing it right now.

PMassicotte (approved):
Looks good to me!

jalvesaq:
The documentation of `vim.treesitter.get_parser()` says:

> Returns the parser for a specific buffer and attaches it to the buffer

So, it seems that each time that `vim.treesitter.get_parser()` is called a new parser is used (although the same one). I will put some `vim.uv.hrtime()` in the code to see if it becomes faster if the parser is stored.

PMassicotte:
> So, it seems that each time that `vim.treesitter.get_parser()` is called a new parser is used (although the same one). I will put some `vim.uv.hrtime()` in the code to see if it becomes faster if the parser is stored.

Curious to see if it would make a significant difference.

jalvesaq:
No difference in a quick test.

jalvesaq:
This is faster than using treesitter (but more complicated): https://github.com/R-nvim/cmp-r/blob/main/lua/cmp_r/init.lua#L328-L396

We are talking about processes that take less than a millisecond, but it's good to know that it's faster to get a range of lines and call `string.find()` on each line than to make a single call to `vim.fn.search()`.

PMassicotte:
This is all good on my side. Ready to merge?

jalvesaq:
Thank you!

PMassicotte:
I really like these changes, code is simpler.

PR #194, opened 2024-07-22
Prevent installing the same missing packages more than once

PMassicotte:
This PR aims at avoiding installation of the same package multiple times. In the current state, it will list `pak1` to be installed twice (if not already installed).

```r
pak1::do_something()
pak1::do_something()
pak2::do_something_else()
```
Basically, this PR just remove duplicate `lintr` messages.

This PR also check for the number of packages to be installed and pluralize the user message accordingly.

## Summary of the changes

✨ (packages.lua): add function to remove duplicate package entries
✨ (packages.lua): add function to create user prompt message for missing packages
♻️ (packages.lua): refactor install_missing_packages to use new functions

### DIFF
--- a/lua/r/init.lua
+++ b/lua/r/init.lua
@@ -8,6 +8,14 @@ M.warn = function(msg)
     )
 end
 
+--- Call vim.notify() with to inform a message
+---@param msg string
+M.inform = function(msg)
+    vim.schedule(
+        function() vim.notify(msg, vim.log.levels.INFO, { title = "R.nvim" }) end
+    )
+end
+
 --- Quick setup: simply store user options
 ---@param opts table | nil
 M.setup = function(opts)
--- a/lua/r/packages.lua
+++ b/lua/r/packages.lua
@@ -1,5 +1,6 @@
 local M = {}
 local S = require("r.send")
+local inform = require("r").inform
 
 --- Removes duplicate entries from a table of packages.
 --- Each package is represented by a table with a 'message' field.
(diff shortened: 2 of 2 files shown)


jalvesaq:
It's better to use `vim.notify()` instead of `print()` because `vim.notify()` can be overridden by user interfaces.
Example of what I see with `vim.notify()`:

```vim
lua vim.notify("This is a message.", vim.log.levels.INFO, { title = "R.nvim" })
```

image


We have the function `require("r").warn()` but perhaps we should create a function `require("r").inform()` and replace some of our warnings with informative messages. The `inform()` function in `lua/r/init.lua` could be:

```lua
--- Call vim.notify() with an informative message
---@param msg string
M.inform = function(msg)
    vim.schedule(
        function() vim.notify(msg, vim.log.levels.INFO, { title = "R.nvim" }) end
    )
end
```

jalvesaq:
Since you are printing this message when there is no diagnostics:

```lua
        print("No missing packages found in the current buffer.")
```

Perhaps you should repeat the same message some lines below if there is no diagnostic of missing packages.

PMassicotte:
I just moved the notify message down.

PMassicotte:
We can wait to merge if you want to create our custom notify function. I will use it here.

jalvesaq:
Function added!

jalvesaq:
Later, I will convert most of the warnings I have added over time into informs...

PMassicotte:
Great!

jalvesaq:
Ready to merge?

PMassicotte:
Yes.

PMassicotte:
Thank you!

Issue #190, opened 2024-07-18
R code lines incorrectly identified


carschandler:
This might be related to the recent change to treesitter identification, but recently, I began having issues where commands that relied on being inside an R code block in a `.Rmd` document stopped working. For example, if I am typing into an empty code block like this:

    ```{r}
    *cursor here*
    ```

and I use the standard `<M-->` bind to enter a `<-`, it doesn't work, but if I am typing above a line of code, i.e.

    ```{r}
    *cursor here*
    a <- 42
    ```

then it works.

I'm experiencing a similar issue where sometimes it will tell me that I'm not over an R-chunk when using `\cc` or `RDSendLine`.

carschandler:
Seems to be any time I'm on the last line of the chunk

PMassicotte:
Probably something off by 1 somewhere.

jalvesaq:
I confirm the bug. It works in Quarto, but not in RMarkdown.

jalvesaq:
Trying to debug this issue, I changed `lualine_x` in my `lualine` config to show the output of `get_lang()`.

```lua
                sections = {
                    lualine_a = { "mode" },
                    lualine_b = { {'branch', icon = ''}, "diagnostics" },
                    lualine_c = { "filename", "searchcount",  },
                    -- lualine_x = { selectionCount },
                    lualine_x = { require("r.utils").get_lang },
                    lualine_y = { {rstatus, color = rsttcolor }},
                    lualine_z = { "progress", "location" },
                },
```

After that, the bug disappeared, even if I removed `get_lang` from my `lualine` config. However, based on the error message that I got in my first attempt to display the output of `get_lang` in the Neovim status line, I created the branch https://github.com/R-nvim/R.nvim/tree/fix_get_lang 

Could you try it and see if it makes any difference (since I could no longer replicate the bug even before the changes)?

jalvesaq:
I updated the branch `fix_get_lang`...

jalvesaq:
It works on my side. So, we might have different tree-sitter parsers. You can try:

- The command `:TSUpdate`
- Add "yaml" to the list of parsers to be installed.
- Delete the `nvim-treesitter` directory to force the installation of new parsers.

jalvesaq:
On the main branch, if I have a chunk of R code in an `Rmd` file with just two lines, `x` and `y`, and press `<M-->` three times after `x` and after `y` I get:

```
x <-  <-  <- 
y <- <M--><M-->
```

[...]

PR #196, opened 2024-07-22
Fix get lang

jalvesaq:
Fix https://github.com/R-nvim/R.nvim/issues/190 (at least in some systems).

### DIFF
--- a/lua/r/send.lua
+++ b/lua/r/send.lua
@@ -543,53 +543,43 @@ end
 M.line = function(m, lnum)
     if not lnum then lnum = vim.api.nvim_win_get_cursor(0)[1] end
     local line = vim.fn.getline(lnum)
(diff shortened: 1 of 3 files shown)


lua/r/send.lua
```diff
@@ -543,53 +543,43 @@ end
...
-            if type(m) == "boolean" and m then
-                knit_child(line, true)
-            else
-                knit_child(line, false)
-            end
-            return
-        end
-    end
-
     local lang = get_lang()
```
PMassicotte, line 546: Nice use of treesitter, this will make things much easier.

PMassicotte (approved):
Looks good to me. I was using the associated branch for the last couple of days and all was good.

jalvesaq:
Thank you!

PR #197, opened 2024-07-26
Call inform() instead of warn() and print()

jalvesaq:
- Call either `inform()` or `warn()` instead of `print()` to print messages.
- Call `inform()` instead of `warn()` when the user doesn't need to fix anything.
- Call `nvim_echo()` instead of `print()` in `RConfigShow`.

### DIFF
--- a/lua/r/config.lua
+++ b/lua/r/config.lua
@@ -112,12 +112,16 @@ local did_real_setup = false
 
 local show_config = function(tbl)
     local opt = tbl.args
+    local out = {}
     if opt and opt:len() > 0 then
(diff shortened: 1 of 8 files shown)


jalvesaq:
@PMassicotte, in this pull request, I replaced some calls to `print()` with either `warn()` or `inform()` in code that you have written. Please, feel free to replace `warn` with `inform` or `inform` with `warn` if you evaluate that it will be better.

PMassicotte (approved):
All good on my side. It's cleaner that way, and your choices between `warn` and `inform` are good ones.

PR #205, opened 2024-08-10
feat(brackets.lua): add support for replacing single brackets with double brackets

PMassicotte:
Basically, will turn `vec[1]` into `vec[[1]]`.


jalvesaq:
Thank you!

PR #207, opened 2024-08-12
Prevent crash when trying to use changing subsetting function in non-R files

PMassicotte:
Due to the absence of Tree-sitter grammar for `.rmd` and `.qmd` files, attempting to use nodes results in a crash. This PR adds a warning to inform the user that the functionality is unavailable outside R files.


jalvesaq:
Thanks!

PR #208, opened 2024-08-12
Do not replace bracket subset matrix

PMassicotte:
This PR will avoid using `[[` on matrix. For example, this will leave this untouched

```r
mat[, , 2]
mat[2 , ,  ]
```


jalvesaq:
Thank you!

PR #215, opened 2024-08-21
Optimize subsetting replacement withing the same node

PMassicotte:
Fix making multiple replacements within the same node, which caused the positions of subsequent replacements to shift, resulting in incorrect text replacements.

For example:

```r
v[a$a, b$b]
```

Was transformed into:

```r
v[a[["a"[["]"]], b$b]
```


jalvesaq:
Thanks!

PR #220, opened 2024-08-26
Optimize subsetting

PMassicotte:
This is a complete rework of this functionality. It avoids using recursive function, which simplify the logic. It is also faster.


jalvesaq:
It's falling in this case:

```r
aa$bb$$cc
```

jalvesaq:
This also doesn't seem to be 100% correct:

```r
aa <- list(bb=c("w" = 1, "x" = 2, "y"= 3, "z" = 4))
aa$bb[3]
```

because the output of `aa$bb[3]` is different from the output of `aa[["bb"]][[3]]`.

PMassicotte:
Good catch. Will work on that.

PMassicotte:
I have re-written this PR. It is now working only for the current diagnostic.

This pull request refactors the `formatsubsetting` function in the `lua/r/format/brackets.lua` file to improve its performance and readability. The changes include removing the use of Tree-sitter queries and simplifying the logic for replacing extraction operators and subsets.

Refactoring and simplification:

* Removed the Tree-sitter query and related parsing logic, which simplifies the code and reduces dependencies.
* Introduced new helper functions `replace_extract_operator` and `replace_subset` to handle specific formatting tasks, making the main function more readable and maintainable.
* Added logic to handle cursor position and diagnostics, ensuring that the function only operates when the cursor is at an appropriate location.

jalvesaq:
I'll take a look at it...

jalvesaq:
It works! Good job! `<LocalLeader>cs` over `$x` converts it into `[["x"]]`, and `["x"]` is converted into `[["x"]]`. This fixes the warnings from https://lintr.r-lib.org/reference/extraction_operator_linter.html

It seems that you should delete this from `doc/R.nvim`:

```
   3. It supports multiple chained subsetting and nested expressions:
         `df$var[1] -> df[["var"]][[1]]
```

PMassicotte:
Thank you very much for your suggestion. I have made the change.

jalvesaq:
Thanks!

PR #223, opened 2024-08-27
fix(config.lua): correct R detection on Windows

### LINKED ISSUE #222: Could not detect R on Windows
I use https://github.com/r-lib/rig to manage R installations on Windows. It writes registry keys to store the R version and path information. R.nvim checks the Windows Registry at `HKEY_LOCAL_MACHINE\SOFTWARE\R-core` to get the path of R installed and set the `config.R_cmd` to `R.exe` if `external_term` is not set.

However, rig creates `R.bat` for `R.exe` (but not `Rgui.bat` and `Rterm.bat` for `Rgui.exe` and `Rterm.exe`), which is directed to the `R.exe` of the default version it sets. This is the content of `R.bat` for me:
```
::4.4.1
@"C:\Program Files\R\R-4.4.1\bin\R" %*
```

This leads to an error when opening R files:

```
Could not find R path in Windows Registry. If you have already installed R, please, set the value of 'R_path'.
Error executing vim.schedule lua callback: Vim:E475: Invalid value for argument cmd: 'R.exe' is not executable
stack traceback:
        [C]: in function 'jobstart'
        ...uanjia/AppData/Local/nvim-data/lazy/R.nvim/lua/r/job.lua:128: in function 'start'
        ...jia/AppData/Local/nvim-data/lazy/R.nvim/lua/r/server.lua:398: in function 'check_nvimcom_version'
        ...jia/AppData/Local/nvim-data/lazy/R.nvim/lua/r/config.lua:851: in function <...jia/AppData/Local/nvim-data/lazy/R.nvim/lua/r/config.lua:851>
```

The simplest solution here is to remove the `.exe` for Line 566. https://github.com/R-nvim/R.nvim/blob/1fb55e5ac50129cebc1d38d88c943a4efe97d60a/lua/r/config.lua#L560-L567

This solution works for me.

hongyuanjia:
### Pull request overview

- Fixes #222 
- Correct the registry query: The root of the Registry entry for R is `HKEY_LOCAL_MACHINE` but not `HKEY_CURRENT_USER.` `HKLM` is also used in Nvim-R. See https://github.com/jalvesaq/Nvim-R/blob/b7cdc9bcd6879465023e20b4a3ae16968edda776/R/windows.vim#L33
- Fix the logic for setting R installation path:
   -  `vim.fn.isdirectory` returns `0` or `1`, but not lua `true` or `false`.
   - `isi386` is a boolean and should not directly compare with `0`

### DIFF
--- a/lua/r/config.lua
+++ b/lua/r/config.lua
@@ -635,27 +635,30 @@ local windows_config = function()
         local reg_roots = { "HKCU", "HKLM" }
         local rip = {}
         for i = 1, #reg_roots do
-            local run_cmd = { "reg.exe", "QUERY", reg_roots[i] .. "\\SOFTWARE\\R-core\\R", "/s" }
-            rip = get_rip(run_cmd)
             if #rip == 0 then
-                -- Normally, 32 bit applications access only 32 bit registry and...
-                -- We have to try again if the user has installed R only in the other architecture.
-                if vim.fn.has("win64") then
-                    table.insert(run_cmd, "/reg:64")
-                else
-                    table.insert(run_cmd, "/reg:32")
-                end
+                local run_cmd = { "reg.exe", "QUERY", reg_roots[i] .. "\\SOFTWARE\\R-core\\R", "/s" }
                 rip = get_rip(run_cmd)
-            end
 
-            if i == #reg_roots and #rip == 0 then
-                warn(
-                    "Could not find R path in Windows Registry. "
-                        .. "If you have already installed R, please, set the value of 'R_path'."
-                )
-                wtime = (uv.hrtime() - wtime) / 1000000000
-                require("r.edit").add_to_debug_info("windows setup", wtime, "Time")
-                return
+                if #rip == 0 then
+                    -- Normally, 32 bit applications access only 32 bit registry and...
+                    -- We have to try again if the user has installed R only in the other architecture.
+                    if vim.fn.has("win64") then
+                        table.insert(run_cmd, "/reg:64")
+                    else
+                        table.insert(run_cmd, "/reg:32")
+                    end
+                    rip = get_rip(run_cmd)
+
+                    if #rip == 0 and i == #reg_roots then
+                        warn(
+                            "Could not find R path in Windows Registry. "
+                                .. "If you have already installed R, please, set the value of 'R_path'."
+                        )
+                        wtime = (uv.hrtime() - wtime) / 1000000000
+                        require("r.edit").add_to_debug_info("windows setup", wtime, "Time")
+                        return
+                    end
+                end
             end
         end
 
--- a/lua/r/config.lua
+++ b/lua/r/config.lua
@@ -631,7 +631,7 @@ local windows_config = function()
             return rip
         end
 
-        local run_cmd = { "reg.exe", "QUERY", "HKCU\\SOFTWARE\\R-core\\R", "/s" }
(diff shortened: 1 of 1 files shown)


PMassicotte (commented):
LGTM.

jalvesaq:
@akthe-at could you check if this pull request works on your side?

akthe-at:
@jalvesaq I will definitely try this when I get to work. I believe I'm the one that moved it away from HKLM because that was not working on Windows for me but the other changes in this PR weren't part of that. Perhaps those will make a differenc, I'll let you all know!

akthe-at:
@jalvesaq @hongyuanjia This looks like it works great! I was concerned about system install vs user level install (I am forced to use a user level install due to my enterprise machine) but the changes in the boolean checks results in this working great. I actually don't even have any issues if I comment out my explicit R_path in my config which used to be an issue for me.

hongyuanjia:
@akthe-at Thanks for confirming it works on your side!

We may need to double-check if it still uses HKLM to store R installation info for user-level installation. Could you further confirm if the R installation path is prepended in your PATH after opening an R file using R.nvim via `:lua print(vim.env.PATH)` ?

akthe-at:
@hongyuanjia 
Checking that shows my R install is on my path. However, upon double checking I do still get the message about `Could not find R path in Windows Registry. If you have already installed R, please, set the value of 'R_path'` if I check :messages (disappeared too fast to notice it before) but I can still start the server and work with the R code just fine. If I add R_path back to my config opts that goes away (and function still remains so this warning seems to be ignorable)

hongyuanjia:
@akthe-at Thanks for the feedback.
Is the R install already on your path before R.nvim loads? That's probably why the code still works; `Rterm.exe` is available regardless of whether it specifies the full path. If this is the case, I would like to add logic to check both `HKLM` and `HKCU`.

akthe-at:
> @akthe-at Thanks for the feedback. Is the R install already on your path before R.nvim loads? That's probably why the code still works; `Rterm.exe` is available regardless of whether it specifies the full path. If this is the case, I would like to add logic to check both `HKLM` and `HKCU`.

/facepalm yup. It is definitely already on my path before that.

hongyuanjia:
@akthe-at Could you please check again if the current approach can remove the warning? I checked HKCU first and then HKLM. Should be able to handle both user-level and system-level installations.

akthe-at:
> @akthe-at Could you please check again if the current approach can remove the warning? I checked HKCU first and then HKLM. Should be able to handle both user-level and system-level installations.

@hongyuanjia Same warning unless I add R_path to config opts...Still functional both ways as well.

hongyuanjia:
Is this because your R's installation process did not write the registry? Could you run the code below to see if there is anything in the locations we are querying?
```r
system2(
    "reg.exe",
    c("QUERY", "HKCU\\SOFTWARE\\R-core\\R")
)

system2(
    "reg.exe",
    c("QUERY", "HKLM\\SOFTWARE\\R-core\\R")
)
```

akthe-at:
image

hongyuanjia:
@akthe-at Thanks! There is a logic error in the previous commit. I hope it will work now...

akthe-at:
@hongyuanjia It is working perfectly now, great work!

hongyuanjia:
@akthe-at Thanks! @jalvesaq I think this PR is ready to merge.

jalvesaq:
Thank you, @hongyuanjia and @akthe-at ! I'll merge it now.

PR #226, opened 2024-08-28
Fix RFormat for entire file

jalvesaq:


jalvesaq:
This fixes `:RFormat` when applied to the entire file, but it only works if using the `styler` package, not `formatR`. It seems that `formatR` can't deal with non-R files such as Quarto documents. Perhaps we should drop the option of using `formatR` and only use `styler`.

jalvesaq:
I removed the integration with the `formatR` package. The pull request is ready to be either squashed and merged  (if we want to keep the `:RFormat` command) or closed (if we expect that other plugins are used for that).

PMassicotte:
Good question. I think we can merge this and refactor/deprecate later if necessary. What do you think?

jalvesaq:
> Good question. I think we can merge this and refactor/deprecate later if necessary. What do you think?

I agree.

PR #228, opened 2024-08-29
auto_start = 1 is not a valid value

idavydov:
replaced invalid `auto_start = 1` with `auto_start = "on startup"` in the example configuration


PMassicotte (approved):

PMassicotte:
Thank you.

PR #240, opened 2024-10-04
Documentation: add some notes about versioning and rework the README a bit

wurli:
As discussed in #237, this PR aims to make it clear to the users when and why breaking changes might occur, and what they can do about it. I've also taken the opportunity to reorganise the README a bit; I've aimed to make it a bit easier to skim-read, and I also added a screenshot at the top.


wurli:
Once this PR is merged I think all we'll need to do is create a 0.1.0 release on GitHub, then it should be possible to merge #237 and bump the version to 0.2.0.

jalvesaq:
> Once this PR is merged I think all we'll need to do is create a 0.1.0 release on GitHub, then it should be possible to merge https://github.com/R-nvim/R.nvim/pull/237 and bump the version to 0.2.0.

I agree, and the PR looks good to me.

I'm unsure if we should put the version number near the top of `doc/R.nvim.txt` as I did in `Nvim-R`. If we do, we must remember to update it with each new release.

wurli:
> If we do, we must remember to update it with each new release.

I *think* shields.io should automatically detect the latest release, so I don't think we'll need to manually change it whenever there's a new release. Currently on [my fork](https://github.com/wurli/R.nvim/tree/versioning) it says 'no releases or repo not found'. I agree though, if for any reason it doesn't automatically update we should change it.

README.md
```diff
@@ -1,199 +1,240 @@
...
+
+## Lifecycle
+
+R.nvim is still maturing and its public API (configuration options,
+commands, and some of the Lua internals) may undergo breaking changes from
+time to time. This project uses [semantic versioning](https://semver.org/) to 
+help with this, and we will always bump the minor version, e.g. from 0.1.x to
+0.2.0, when we make a breaking change. Users are thus encouraged to pin their
+installation of R.nvim to the **latest minor release** and to check the release
+notes for any breaking changes when upgrading.
```
PMassicotte, line 139: Breaking changes should be when we change major, not minor right?

README.md
```diff
@@ -1,199 +1,240 @@
...
+    components concatenated using `paste()` (requires
+    [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter) to be
+    installed).
 
-There are two new commands available to separate a file path into its different
-components.
+- `<LocalLeader>sh` splits a filepath under the cursor into individual 
+    components concatenated using `here::here()` (requires
+    [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter) to be
+    installed).
```
PMassicotte, line 211: This is now only 1 function `<localleader>sp`. The split function to use can be configured with `path_split_fun` which can be 

```lua
  path_split_fun   = { "here::here", "here", "file.path", "fs::path", "path" }
```

README.md
```diff
@@ -235,15 +276,19 @@ but temporary files are used in a few cases.
...
+- [cmp-r](https://github.com/R-nvim/cmp-r): autocompletion source for
+    [nvim-cmp](https://github.com/hrsh7th/nvim-cmp) using R.nvim as backend.
+
+- [languageserver](https://github.com/REditorSupport/languageserver): a
+    language server for R.
+
+- [colorout](https://github.com/jalvesaq/colorout): a package to colorize R's
+    output.
+
+- [Ark](https://github.com/posit-dev/ark): a LSP server/DAP server/Jupyter kernal
```
PMassicotte, line 288:
```suggestion
- [Ark](https://github.com/posit-dev/ark): a LSP server/DAP server/Jupyter kernel
```

PMassicotte (commented):
Other than that, it looks good to me. Thank you!

README.md
wurli: Semver allows breaking changes before 1.0.0, but not after:

> 4. Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.
>
> 5. Version 1.0.0 defines the public API. The way in which the version number is incremented after this release is dependent on this public API and how it changes.

README.md
wurli: Fixed!

PMassicotte (approved):

PMassicotte:
All good on my side!

wurli:
Thanks! 🚀

PR #245, opened 2024-10-10
when using RbrowseURLfun don't call open cli command, fixes #232

### LINKED ISSUE #232: open_html doesn't work if getOption("browser") is a function
I noticed that if I set `options(browser=function(url) {...})` then `open_html` setting is ignored.

I think it's a valid option and is correctly used e.g. by `browserURL()`.

Would it maybe be possible to allow browser which is a function? I am using this to view html's from a remote host using a combination of ssh port forwarding, servr and opener. 

Not sure if my solution is optimal, by happy to document it in a wiki.

idavydov:


jalvesaq:
I fixed a bug in your pull request. It should be OK now.

jalvesaq:
Merged now. Thank you!

PR #246, opened 2024-10-10
Refactor Codebase for Improved Modularity and Asynchronous Operations (WIP)

she3o:
This pull request introduces several refactors aimed at enhancing the modularity and performance of the Lua source. The changes are focused on organizing functionalities into dedicated modules, centralizing common operations, and leveraging asynchronous capabilities where it's worth the callback hell.

# Summary of Changes

1. Moved logging functionality to dedicated `log.lua`
	- Benefits:
		- For new potential contributors, `log.lua` is a more expected place to find `warn` and `inform` than `init.lua`.

2. Moved user commands registration to dedicated `commands.lua`
	- Benefits:
		- `config.lua` does a lot of thing. Many of which are not related to applying configuration.
		- Separation of concerns: command definitions and configuration.
		- Makes the codebase more scalable for future additions.

3. Moved some local functions in `config.lua`
	- Moved `resolve_fullpaths` from `config.lua` to `utils.lua`.
	- Some tables and functions are only used in User commands. I moved them to `commands.lua`. I might move them again later to a more appropriate location.

4. Other improvements
	- Added `check_executable` Function in `utils.lua`:
		- Verifies executables like `vim.fn.executable` but async.
		- TODO: currently it only runs commands with `--version`. Ideally it should also handle commands without `--version` or those that don't exit after running with `--version`.
	- Replaced `vim.fn.executable` with `check_executable` in `config.lua` and `server.lua`.
	- Switched from vim.loop to vim.uv in some files. Later I will do it everywhere.
	- Refactored the `R CMD build` process for building `nvimcom` into a separate `build_package` function.
	- Benefits:
		- reduce potential blocking operations.
		- I will introduce more async enhancements hopefully without introducing bugs.

# Rationale

- By breaking down the codebase into focused modules (`log.lua`, `commands.lua`, `utils.lua`) that have self-explanatory names and separate concerns, we make it easier for new contributors to navigate and understand the code.
- Centralizing common functionalities reduces code duplication.
- Make the codebase more adaptable to future requirements (i.e. new features and optimizations).
-


PMassicotte:
I checked out the PR and after a rapid reading it looks good. I'll test it later today to see if there are no issues (would be surprising). Thanks for the PR!

jalvesaq:
Thnaks, @she3o!

Some comments:

- In `doc/R.nvim.txt` we say that we support Neovim 0.9.5, which is the version on Debian testing and Ubuntu LTS. On Debian testing either `:lua print(vim.uv)` or `:lua print(vim.loop)` prints the address of the same table. So, it seems that we already can completely eliminate any reference to `vim.loop` (`grep -R 'vim.loop' lua` reveals `vim.loop` called sometimes).
- The function `utils.check_executable()` is called four times. In all four, a warning message is emitted when the executable isn't found. So, the signature of `utils.check_executable()` could be simplified to `check_executable(cmd, msg)`. In `server.lua`, it would be:
```
if check_executable("R", '"R" executable not found.') then
    build_package()
end
```

lua/r/log.lua
PMassicotte, line 1: Maybe simplify like this:

```lua
local M = {}

--- Helper function to call vim.notify() with scheduled execution
---@param msg string
---@param level number
---@param opts table
local function notify(msg, level, opts)
    vim.schedule(function() vim.notify(msg, level, opts) end)
end

--- Call vim.notify() with a warning message
---@param msg string
M.warn = function(msg) notify(msg, vim.log.levels.WARN, { title = "R.nvim" }) end

--- Call vim.notify() to inform a message
---@param msg string
M.inform = function(msg)
    notify(msg, vim.log.levels.INFO, { title = "R.nvim", hide_from_history = true })
end

return M
```

PMassicotte (changes requested):
Looks good to me, just small suggestions.

wurli:
One additional suggestion from me which feels like it may be in the spirit of this PR. Perhaps we should create an additional `api.lua` module where we alias all functions we currently consider to be 'public', i.e. ones which users can rely on. Currently this list only includes `r.run.action()` and `r.send.cmd()`. I think this would make 'public' functions easier for devs and users to keep track of if we add more in future. Since there are only 2 functions in the list right now, a change at this stage would be more about organisation/futureproofing than anything else.

jalvesaq (approved):

jalvesaq:
I didn't see any new bugs introduced by this PR. So, it looks good to me.

jalvesaq:
> Perhaps we should create an additional api.lua module where we alias all functions we currently consider to be 'public', i.e. ones which users can rely on.

@wurli, I like this idea.

PR #253, opened 2024-10-23
Add lua/r/health.lua

### LINKED ISSUE #251: Not inside R or Python code chunk.
For .Rmd files, a similar behaviour as described in https://github.com/R-nvim/R.nvim/issues/190 occurs. 
I.e. a new .Rmd file with 
```{r}
a <- 1 *cursor*
```
gives "Not inside R or Python code chunk".
Temporarily, :InspectTree fixes this, but this does not work for long. 

Modifying line 577 of r/send.lua to 

if lang ~= "r" and lang ~= "markdown" then

seems to fix this.

jalvesaq:
Close #251

### DIFF
--- a/lua/r/health.lua
+++ b/lua/r/health.lua
@@ -1,0 +1,69 @@
+--- Adapted from markview.nvim plugin
+
+local M = {}
+
+local ts_available, treesitter_parsers = pcall(require, "nvim-treesitter.parsers")
+
+--- Checks if a parser is available or not
+---@param parser_name string
+---@return boolean
+local function parser_installed(parser_name)
+    return (ts_available and treesitter_parsers.has_parser(parser_name))
+        or pcall(vim.treesitter.query.get, parser_name, "highlights")
+end
+
+M.check = function()
+    vim.health.start("Checking applications and plugins:")
+
+    if vim.fn.has("nvim-0.9.5") == 1 then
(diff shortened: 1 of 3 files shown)


jalvesaq:
The health module is called by Neovim even if not editing an R file. This means that we can't check the validity of user options such as  `R_app`. I didn't add a check for `tmux` because it's possible to run R in an external `kitty` terminal emulator without `tmux`. Should you check if `tmux` is installed anyway? Should we add any other test to the module?

lua/r/health.lua
```diff
@@ -0,0 +1,69 @@
+--- Adapted from markview.nvim plugin
+
+local M = {}
+
+local ts_available, treesitter_parsers = pcall(require, "nvim-treesitter.parsers")
```
PMassicotte, line 5: Should it be moved inside the check() function?

PMassicotte:
Very nice idea. we should have had this long time ago, good jog.

> Should you check if `tmux` is installed anyway?

Maybe just a warning?

> Should we add any other test to the module?

Good question. I can not think of anything else at the moment.

PMassicotte:
Looks good to me for merging now.

jalvesaq:
Thanks! I merged it, maybe too quickly... I was also thinking that Tmux could be added as a warning, but we can add more diagnostics in the future.

PR #278, opened 2024-11-23
Add type hints for user config

wurli:
This PR doesn't add any new functionality, but adds type definitions and hints for the full set of configuration options which are available to the user through `setup()`. This is purely a quality-of-life thing; if a user is using `lua_ls` as their language server for Lua, they should now receive live feedback if they try to enter any invalid configuration during setup, and they'll also get brief hover-info for each of the available options.

Also:

* Fixed some typos in the documentation.
* Removed default options `help_w` and `non_r_compl` which weren't being used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,13 +53,14 @@ A longer example adding some custom behaviour:
     version = "~0.1.0"
     config = function()
         -- Create a table with the options to be passed to setup()
+        ---@type RConfigUserOpts
         local opts = {
             hook = {
                 on_filetype = function()
                     vim.api.nvim_buf_set_keymap(0, "n", "<Enter>", "<Plug>RDSendLine", {})
                     vim.api.nvim_buf_set_keymap(0, "v", "<Enter>", "<Plug>RSendSelection", {})
                 end
-            }
+            },
             R_args = {"--quiet", "--no-save"},
             min_editor_width = 72,
             rconsole_width = 78,
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -1041,8 +1041,8 @@ put in your config:
 >lua
    rmdchunk = "``"
 <
-If `rmdchunk = 2` (the default), in addition of creating an empty chunk
-block, a single backtick will be converted in inline R delimiters if the
+If `rmdchunk = 2` (the default), in addition to creating an empty chunk
+block, a single backtick will be converted to inline R delimiters if the
 cursor is not at the beginning of the line.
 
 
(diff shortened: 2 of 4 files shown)


PMassicotte:
LGTM.

jalvesaq:
It seems great, but I have `lua_ls` and still can't see any diagnostics when entering invalid configuration options. Do I need a specific configuration for `lua_ls`?

PMassicotte:
I just tested and have the same question as @jalvesaq

wurli:
That's odd 🤔 Here's a gist with a minimal `init.lua` that gives me type hints: 

<script src="https://gist.github.com/wurli/2256d7410f2cec4fe9f35ee7e0316056.js"></script>

With some testing I found that if I opened the script and the line `require("r").setup({})` was already present I wouldn't get any type hints. If I deleted the line and re-added it I'd get hints though; I guess actually writing the line is what tells `lua_lsp` to unpack type hints from the module.

jalvesaq:
Great! It works with your minimal `init.lua`!

I still have to find what's missing in my Neovim configuration.

jalvesaq:
It doesn't work if I do this:

```lua
	local opts = {
	}
	require("r").setup(opts)
```

And I need the `opts` table because I have some conditional options:

```lua
            if vim.o.columns > 200 then
                opts.rconsole_height = 14
                opts.objbr_auto_start = true
                opts.objbr_place = "script,left"
                opts.objbr_w = 60
                opts.rconsole_height = 24
            elseif vim.o.columns > 140 then
                opts.rconsole_height = 20
                opts.objbr_w = 35
            else
                opts.rconsole_height = 14
                opts.objbr_w = 25
            end
```
```

**Comment by wurli** [comment]
@jalvesaq I see - that makes sense! In such cases you should be able to force the LSP to apply the class like so:

``` lua
---@type RConfigUserOpts
local opts = {}
```

jalvesaq:
Thanks! It works with your `init.lua`. I still have to figure out how to fix my configuration, and I don't have time now, but the pull request looks good.

There is a warning about a duplicate field in `lua/r/config.lua`, but it can be fixed by:

```diff
diff --git a/lua/r/config.lua b/lua/r/config.lua
index f8c121c..d1c47bb 100644
--- a/lua/r/config.lua
+++ b/lua/r/config.lua
@@ -305,7 +305,7 @@ local hooks = require("r.hooks")
 ---S
 ---Whether to replace `<` with `<<>>=\n@` when writing Rnoweb files; defaults
 ---to `true`. See |rnowebchunk| or `:help rnowebchunk` for more information.
----@field rnowebchunk         = true,
+---@field rnowebchunk? boolean
 ---
 --The directory containing R.nvim's plugin files.
 --@field private rnvim_home? string
```

My original Neovim config marks all fields as duplicates, but I have to fix this in my config.

wurli:
Good catch! Fixed and will merge now. Thanks both for review :)

PMassicotte:
Thank you!

PR #282, opened 2024-11-28
expand ~ in server.lua

Jacques-z:
Installing this plugin on my new PC who had just installed R, I fell into repeated prompt for `mk`ing`dir`.
'Cause it made a dir `./~/R/...`. Fixed by adding `expand`.


PMassicotte:
LGTE.

PR #286, opened 2024-12-06
set params in global env & introduce set_params option, fixes #285

### LINKED ISSUE #285: cannot change value of locked binding for 'params'
In cases when a library has an object names `params` (e.g., `S4Vectors`) the logic of assigning `params` breaks:

```
Error in params <<- old_params : 
  cannot change value of locked binding for 'params'
```

To reproduce, create an `.Rmd` file:
````
---
params:
  a: 2
---

```{r}
library(S4Vectors)
```

```{r}
params$a
```
````  

Open it in nvim and render to html `'kh`:
```
> nvim.interlace.rmd("a.Rmd", outform = "html_document", rmddir = "<...>/tmp/omparams", envir = .GlobalEnv)


processing file: a.Rmd
                                                                              
                            
output file: a.knit.md

/opt/conda/bin/pandoc <...> /tmp/RtmpPPCldU/rmarkdown-str29ee3e876560.html 
[WARNING] This document format requires a nonempty <title> element.
  Defaulting to 'a.knit' as the title.
  To specify a title, use 'title' in metadata or --metadata title="...".

Output created: a.html
Error in params <<- old_params : 
  cannot change value of locked binding for 'params'
```

### PR DESCRIPTION
**Pull request by idavydov**
- More robust handling of environments when updating `params`
- Introduce `set_params` config option; setting `set_params="no"` disables the feature. The reasoning is that modifying global environment is not always desirable, so a user might want to control this behavior. 

I made `set_params` a string, as I can imagine there could be more behaviors. E.g., maybe `set_params="no_override"` is useful, when `R.nvim` will set the variable initially but never override it if it is set already.

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -747,6 +747,7 @@ command:
 |open_html|           Open HTML after processing Rmd or Quarto
 |insert_mode_cmds|    Allow R commands in insert mode
 |rmhidden|            Remove hidden objects from R workspace
+|set_params|          Control setting `params` in `.GlobalEnv` (`"no" ` to disable).
 |wait|                Time to wait for nvimcom loading
 |wait_reply|          Time to wait for R reply
 |setwd|               Directory where to start R
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -747,7 +747,7 @@ command:
 |open_html|           Open HTML after processing Rmd or Quarto
 |insert_mode_cmds|    Allow R commands in insert mode
 |rmhidden|            Remove hidden objects from R workspace
-|set_params|          Control setting `params` in `.GlobalEnv` (`"no" ` to disable).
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by jalvesaq** [comment]
I added the option "no_override" to `set_params`. Please, do a `git pull` in your `fix285` branch and do this to try the new changes:

```
echo "remove.packages('nvimcom')" | R
```

jalvesaq:
I agree.

jalvesaq:
I moved the config condition to a position after the R_Nvim_status because it seems more logical. But this is a cosmetic change with no real difference.

jalvesaq:
Can we merge it now?

idavydov:
I did not test extensively (especially `no_override`), but if prefer to merge early please go ahead. If something is not working as expected I'll make a ticket and/or PR.

jalvesaq:
In a quick test, it works. Merging now... Thank you!

PR #290, opened 2024-12-10
Fix wrong resolve of argument completion

jalvesaq:
If the argument name appears in the description of a previous argument preceded by a empty space and followed by a comma, cmp-r would show the wrong description for the argument. For, example, when resolving the "habillage" argument from the plot.MCA function (FactoMineR package).

### DIFF
--- a/nvimcom/DESCRIPTION
+++ b/nvimcom/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: nvimcom
-Version: 0.9.59
-Date: 2024-12-06
+Version: 0.9.60
+Date: 2024-12-10
 Title: Intermediate the Communication Between R and Neovim
 Author: Jakson Aquino
 Maintainer: Jakson Alves de Aquino <jalvesaq@gmail.com>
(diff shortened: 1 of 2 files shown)


jalvesaq:
To replicate the bug fixed by this pull request:

- Install the package `FactoMineR`.
- Create an R script with the following code:

```r
library(FactoMineR)
plot.MCA(habill)
```

- In Insert mode, add the letter "a" to "habill" to trigger the completion of "habillage" and select the item "habillage" from the completion menu.

Expected result:

```
habillage: string corresponding to the color which are used.
If "none", one color is used for the individual, another one
for the categorical variables; if "quali", one color is used
for each categorical variables; else if it is the position of
a categorical variable, it colors according to the different
categories of this variable
```

Wrong result:

```
col.quanti.sup^Ecol.quanti.sup: a color for the supplementary
quantitative variables, if color ="none" the label is not
written
```

PMassicotte:
I can replicate the issue and that PR fixes it. Thank you!

PR #293, opened 2024-12-16
Check if current node is "program"

### LINKED ISSUE #221: Sending code to console moves the cursor down two lines, not one
This issue arose without making any changes to my configuration, and only occurs on MacOS, not linux (Ubuntu). It has been an issue ever since I started using MacOS for work.

Here is an example of what happens. In this example I'm pushing <CR> from the first line while in normal mode.

```r
sqrt(16)   <-- this line is executed
sqrt(4)     <-- this line is skipped
sqrt(36)   <-- the cursor ends up here
```

The cursor ends up on the third line (and, as expected, only the first line was executed). This happens no matter how many lines are in the file (except if it only has one line). 

I would try to look into this myself but I'm not sure where to start.

jalvesaq:
Close #221

### DIFF
--- a/lua/r/send.lua
+++ b/lua/r/send.lua
@@ -114,6 +114,8 @@ local function get_code_to_send(txt, row)
         ignore_injections = false,
     })
 
+    if node and node:type() == "program" then node = node:child(0) end
+
     while node do
         local parent = node:parent()
         if


jalvesaq:
This is a workaround to a tree-sitter problem: when the cursor is on the first child node, `nvim-treesitter` identifies the node as the root node ("program") and not the first child node.

I quickly tested the changes with R and Quarto and found no problem.

PMassicotte:
LGTM. It should not cause any problem when/if the issue I reported on nvim is fixed.

PR #295, opened 2024-12-21
Set params when loading nvimcom

### LINKED ISSUE #294: R terminal crash on qmd  open or  leader RF
Hi 

In the last couple of days I have been having difficulties working with quarto files on windows.

When I select a QMD file and then initiate an R terminal the following error is returned:

Warning: stack imbalance in '::', 71 then 69
Warning: stack imbalance in '$', 68 then 66 
[Process exited -1073741784]

If I first open an R file I am able to initialise the R terminal.

But when I then open a Quarto file I get the following error

> Error: '\R' is an unrecognized escape in character string (<input>:1:29)

[Process exited -1073741784]

I am happy to provide additional information if you can let me know what would be useful.

Many thanks.

jalvesaq:
Close #294

### DIFF
--- a/lua/r/rmd.lua
+++ b/lua/r/rmd.lua
@@ -147,7 +147,40 @@ M.next_chunk = function()
     end
 end
 
-local has_params = false
+local last_params = ""
+
+--- Check if the YAML field params exists and if it is new
+--- @return string
+M.params_status = function()
+    local lines = vim.api.nvim_buf_get_lines(0, 0, -1, true)
+    if lines[1] ~= "---" then
+        if last_params == "" then return "unchanged" end
+        last_params = ""
+        return "deleted"
+    end
+
+    local i = 2
+    while i < #lines do
+        if lines[i] == "params:" then
+            local cp = ""
+            i = i + 1
+            while i < #lines and lines[i]:sub(1, 1) == " " do
+                cp = cp .. lines[i]
+                i = i + 1
+            end
+            if last_params == cp then return "unchanged" end
+            last_params = cp
+            return "new"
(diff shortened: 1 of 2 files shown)


jalvesaq:
This pull request sets the value of the params object while loading nvimcom to avoid a crash on Windows.

jalvesaq:
@jasonpott, the R code to create the params object now is called only when there is any change in the params field which is safer than running it whenever the document is saved.

@idavydov may also want to review this pull request...

lua/r/rmd.lua
```diff
@@ -147,47 +147,64 @@ M.next_chunk = function()
...
         if lines[i] == "params:" then
-            p = true
-            break
+            local cp = ""
+            i = i + 1
+            while i < #lines and lines[i]:sub(1, 1) == " " do
+                cp = cp .. lines[i]
+                i = i + 1
+            end
+            if last_params == cp then return "old" end
```
PMassicotte, line 171: Is `old` ever used elsewhere?

jalvesaq:
> Is old ever used elsewhere?

No. It has the same practical effect as "none": _Don't run any R code_. However, "none" wouldn't correspond to the reality when there is already an old params field. Perhaps we could return "unchanged" for both cases.

PMassicotte:
Other than that, I tried on my Linux machine and it works.

PMassicotte:
Thanks for this @jalvesaq

jalvesaq:
> If it would be helpful I'm happy to test future versions on windows if that would help the project?

Yes, thank you!

PR #297, opened 2024-12-24
Display `:RMapsDesc` and `:RConfigShow` in a split buffer instead of as an `nvim_echo()` statement

kkanden:
This change will make outputs of these two commands display in a separate buffer, which will allow for easier navigation and searching for specific commands, mappings and config options. This will also mimic the behavior of `:h` making it less dizzying.

### DIFF
--- a/lua/r/commands.lua
+++ b/lua/r/commands.lua
@@ -3,74 +3,74 @@ local config = require("r.config").get_config()
 local M = {}
 
 local function show_config(tbl)
-	local opt = tbl.args
-	local out = {}
-	if opt and opt:len() > 0 then
-		opt = opt:gsub(" .*", "")
-		table.insert(out, { vim.inspect(config[opt]) })
-	else
-		table.insert(out, { vim.inspect(config) })
-	end
+    local opt = tbl.args
+    local out = {}
+    if opt and opt:len() > 0 then
+        opt = opt:gsub(" .*", "")
+        table.insert(out, { vim.inspect(config[opt]) })
+    else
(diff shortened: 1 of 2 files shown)


jalvesaq:
Thanks for your contribution! Please, create a symbolic link to `scripts/pre-commit` as explained in `CONTRIBUTING.md` and run `stylua`:

```
cd /path_to/R.nvim
stylua lua
```

jalvesaq:
Maybe we could add this to the buffers (as we do for R documentation):

```lua
    vim.api.nvim_buf_set_keymap(0, "n", "q", ":q<CR>", { noremap = true, silent = true })
```
Perhaps we shouldn't use a buffer when an argument is passed to `:RConfigShow` because a single value is displayed. Example:

```
:RConfigShow compldir
```

Although it might be a table:

```
:RConfigShow compl_data
```

What do you think about these suggestions?

kkanden:
I think those are good ideas. I forgot that you could pass na argument to `:RConfigShow`. Maybe if an argument is passed, the value of that option could just be displayed using `nvim_echo()` as it is now, and if no argument is passed, only then display it in a buffer?

jalvesaq:
> Maybe if an argument is passed, the value of that option could just be displayed using nvim_echo() as it is now, and if no argument is passed, only then display it in a buffer?

Yes, I agree.

kkanden:
I've implemented the suggestions. I have also added highlighting to the keywords in `:RConfigShow` for better looks :)

jalvesaq:
Thank you! It was a good improvement to the two commands!

PR #299, opened 2024-12-28
Fix sending code to R from Rnoweb documents

### LINKED ISSUE #298: Rnoweb: Error: unexpected input in "<<"
image

Thank you very much for R.nvim at all the work done. I'm using r.nvim and the same configuration for rmd files without problems, but in rnoweb files I'm observing the following behavior.

When I send the **first** line after a chunk , see line 10 (via 	vim.api.nvim_buf_set_keymap(0, "n", "<Space>", "<Plug>RDSendLine", {}) ) i have an error message, but the next one (line 11) can be sent without error message.
The command is correctly executed and i can also knit the pdf correctly. It's not a real bug.

I tried also with minimal configuration. I had the same "error". 

I also remarked that in rmd  the autocomplete for chunk options is working, this is not the case for rnoweb files. I don't know if this maybe related.

:set filetype? -> rnoweb (filetype is recognized)

I added roman2.txt which is indeed a rnoweb file (roman2.rnw).

[roman2.txt](https://github.com/user-attachments/files/18261070/roman2.txt)

Can you reproduce this or do you have a solution.

Thanks

jalvesaq:
Close #298

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -88,7 +88,7 @@ Main dependencies:
    nvim-treesitter:
       https://github.com/nvim-treesitter/nvim-treesitter
       You have to enable highlighting by tree-sitter and install parsers for
-      `r`, `markdown`, `markdown_inline`, `rnoweb`, `yaml`, and `csv`.
+      `r`, `markdown`, `markdown_inline`, `rnoweb`, `yaml`, `latex`, and `csv`.
 
    cmp-r (for auto-completion):
       https://github.com/R-nvim/cmp-r
@@ -103,27 +103,29 @@ Main dependencies:
       On Windows, you have to install Rtools to be able to build `nvimcom`:
       https://cran.r-project.org/bin/windows/Rtools/
 
-   Note about the R package `nvimcom`:
+      Note about the R package `nvimcom`:
 
-     - You do not need to load nvimcom in your .Rprofile because the R.nvim
-       plugin sets the environment variable `R_DEFAULT_PACKAGES`, including
-       `nvimcom` in the list of packages to be loaded on R startup. However,
-       if you set the option `defaultPackages` in a .Rprofile, you should
-       include "nvimcom" among the packages to be loaded (see
-       |nvimcom-not-loaded|).
+      You do not need to load nvimcom in your .Rprofile because the R.nvim
+      plugin sets the environment variable `R_DEFAULT_PACKAGES`, including
+      `nvimcom` in the list of packages to be loaded on R startup. However,
+      if you set the option `defaultPackages` in a .Rprofile, you should
+      include "nvimcom" among the packages to be loaded (see
+      |nvimcom-not-loaded|).
+
+   Quarto: required to complete chunk options in `#|` comments in Quarto, Rmd
+      and Rnoweb documents: https://quarto.org/
 
 
 Additional dependencies for editing Rnoweb documents:
 
-   latexmk:       Automate the compilation of LaTeX documents.
-                  See examples in |latexcmd|.
+   latexmk: Automate the compilation of LaTeX documents.
+            See examples in |latexcmd|.
 
    PDF viewer with SyncTeX support and capable of automatically reloading
    documents. This is required only if you edit Rnoweb files.
       On Linux and other Unix systems: Zathura.
       On macOS: Skim.
       On Windows: SumatraPDF.
-
 
 Suggestions for Unix (Linux, macOS, etc.):
 
@@ -345,7 +347,7 @@ To get help on an R topic, you can also type in Neovim (Normal mode):
 >vim
    :RHelp topic
 <
-You may close the R documentation buffer by simply pressing `q`. R
+You may close the R documentation window by simply pressing `q`. R
 documentation is open in a 'modifiable' buffer to allow you to edit the
 code in the "Examples" section. Press `<LocalLeader>`gn to jump to the
 next section of R code ("Usage" or "Examples" section) and use R.nvim
@@ -394,6 +396,8 @@ works if R is running. The value of 'shiftwidth' is passed as `indent_by`.
 
 ------------------------------------------------------------------------------
 4.2. Editing Rnoweb files
+
+See |rnoweb-chunks| to know how to properly include chunk options.
 
 In Rnoweb files (.Rnw), when the cursor is over the `@` character, which
 finishes an R chunk, the command to send a line to R Console will make cursor
@@ -621,7 +625,30 @@ until an empty line is found.
 
 
 ------------------------------------------------------------------------------
-5.5 You can't put `syntax enable` in your `init.vim` (macOS only)
+5.5 Old style chunks not properly recognized in Rnoweb         *rnoweb-chunks*
+
+Chunks of R code with old style headers are not correctly recognized by
+tree-sitter and, consequently, it's not properly highlighted and sent to the
+R Console. So, instead of:
+>
+   <<example, echo=TRUE, results="asis">>=
+   x <- 1
+   print(x)
+   @
+<
+You should write:
+>
+   <<example>>=
+   #| echo: true
+   #| results: "asis"
+
+   x <- 1
+   print(x)
+   @
+<
+
+------------------------------------------------------------------------------
+5.6 You can't put `syntax enable` in your `init.vim` (macOS only)
 
 It seems that if you put the command `syntax enable` in your `init.vim` on OS
 X, file type plugins are immediately sourced. Consequently, all R.nvim
@@ -633,7 +660,7 @@ details, please, access:
 
 
 ------------------------------------------------------------------------------
-5.6. Windows bugs and limitations
+5.7. Windows bugs and limitations
 
 ------------------------------
 Wrong message that "R is busy"
@@ -687,6 +714,8 @@ command:
 >vim
    :RConfigShow
 <
+You may close the window opened by `:RConfigShow` by simply pressing `q`.
+
 |auto_start|          Start R automatically
 |objbr_auto_start|    Start the Object Browser automatically
 |built-in-terminal|   Options to control Neovim's built-in terminal
@@ -1937,7 +1966,7 @@ Use `data.table` to save a TSV file:
 >lua
    view_df = {
      n_lines = 0,
-     save_fun = "function(obj, obj_name) {f <- paste0(obj_name, '.tsv'); data.table::fwrite(obj, f, sep = '\t') ; f}",
+     save_fun = "function(obj, obj_name) {f <- paste0('/tmp/', obj_name, '.tsv'); data.table::fwrite(obj, f, sep = '\t') ; f}",
    }
 
 Use `arrow` to save a parquet file, and VisiData to open it:
@@ -2115,7 +2144,8 @@ mapped in the current session, do in Normal mode:
 >vim
    :RMapsDesc
 <
-The prefix "RD" in the labels means "cursor down".
+You may close the window opened by `:RMapsDesc` by simply pressing `q`. The
+prefix "RD" in the labels means "cursor down".
 
 The plugin contains a function called `action` in the Lua module `r.run`
 which allows you to build ad-hoc commands to R. This function takes the name
--- a/lua/r/send.lua
+++ b/lua/r/send.lua
@@ -757,7 +757,7 @@ M.funs = function(bufnr, capture_all, move_down)
 
     bufnr = bufnr or vim.api.nvim_get_current_buf()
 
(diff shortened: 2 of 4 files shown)


jalvesaq:
This pull request fixes an issue by calling a function that does nothing:

```lua
    if node then vim.schedule(function() end) end
```

Clearly, this shouldn't be necessary and we are circumventing what seems to be a Neovim or tree-sitter bug. I don't know whether the fix works or introduces new bugs on macOs or Windows. So, the pull request needs review.

jalvesaq:
Example of Rnoweb document that can be used to test the changes:

`Example.Rnw`:

```rnw
\documentclass{article}

\begin{document}

Here is a code chunk.

<<example>>=
x <- 1
print(x)
x |>
    print()
@

And another one:

<<example2>>=
#| echo: false
#| fig-cap: "Example of Figure"

# Cursor on `a`. Send the line and jump to `b`.
a <- 1
b <- 2

# Cursor on `d`. Send the line and jump to `if`.
d <- data.frame(var1 = c(1, 2, 3, 4, 5), var2 = c(1, 2, 3, 3, 2))

# Cursor on `if`. Send the next two lines.
if (TRUE)
    library("ggplot2")

# Cursor on `if`. Send the bracket block.
if (TRUE) {
    # Cursor on `print`. Send the line.
    print(1)
    print(2)
}

# Cursor on `plus_one`. Send the bracket block.
plus_one <- function(a) {
    # Cursor on `b`. Send the line.
    b <- a + 1
    # <LocalLeader>fc here. Send the function `plus_one`.
    nested_fun <- function(x) {
    # <LocalLeader>fc here. Send the function `plus_one`.
        y <- x + 1
        return(invisible(y))
    }
    b
}

# <LocalLeader>fa. Send the functions `plus_one` and `plus_two`.

# Cursor anywhere on the line below. Send the whole line and jump to `d`.
a <- 1 ; b <- 2
d <- "a;b"

library("dplyr")
library("ggplot2")

# Statements chained by a pipe operator
# Cursor anywhere in the three lines below. Send the three lines.
mtcars |>
    select(mpg, qsec) |>
    cor()

# Statements chained by arithmetic operator
# Cursor anywhere in the three lines below. Send the three lines.
ggplot(mtcars, aes(mpg, qsec)) +
    geom_point() +
    geom_smooth(span = 3.0)
@

End of document.


\end{document}
```

PMassicotte:
Testing it now...

PMassicotte:
In your `example.Rnw`, at the 

```r
# <LocalLeader>fa. Send the functions `plus_one` and `plus_two`.
```

I am getting `Not in an R file` when I try to execute it.

doc/R.nvim.txt
```diff
@@ -103,28 +103,30 @@ Main dependencies:
...
-       include "nvimcom" among the packages to be loaded (see
-       |nvimcom-not-loaded|).
+      You do not need to load nvimcom in your .Rprofile because the R.nvim
+      plugin sets the environment variable `R_DEFAULT_PACKAGES`, including
+      `nvimcom` in the list of packages to be loaded on R startup. However,
+      if you set the option `defaultPackages` in a .Rprofile, you should
+      include "nvimcom" among the packages to be loaded (see
+      |nvimcom-not-loaded|).
+
+   Quarto: required to complete chunk options in `#|` comments in Quarto, Rmd
```
PMassicotte, line 115:
```suggestion
   Quarto: the `quarto` R package is required to complete chunk options in `#|` comments in Quarto, Rmd
```

doc/R.nvim.txt
```diff
@@ -621,7 +625,30 @@ until an empty line is found.
 
 
 ------------------------------------------------------------------------------
-5.5 You can't put `syntax enable` in your `init.vim` (macOS only)
+5.5 Old style chunks not properly recognized in Rnoweb         *rnoweb-chunks*
+
+Chunks of R code with old style headers are not correctly recognized by the
```
PMassicotte, line 630:
```suggestion
Chunks of R code with old style headers are not correctly recognized by
```

doc/R.nvim.txt
```diff
@@ -1937,7 +1966,7 @@ Use `data.table` to save a TSV file:
 >lua
    view_df = {
      n_lines = 0,
-     save_fun = "function(obj, obj_name) {f <- paste0(obj_name, '.tsv'); data.table::fwrite(obj, f, sep = '\t') ; f}",
+     save_fun = "function(obj, obj_name) {f <- paste0('/tmp/', obj_name, '.tsv'); data.table::fwrite(obj, f, sep = '\t') ; f}",
```
PMassicotte, line 1969: Is this just to fix the documentation?

doc/R.nvim.txt
jalvesaq: We need the completion data from the quarto _application_ for Rmd, Quarto, and Rnoweb. The quarto _package_ is required only to compile Quarto documents directly from R.

doc/R.nvim.txt
```diff
@@ -2115,7 +2144,8 @@ mapped in the current session, do in Normal mode:
 >vim
    :RMapsDesc
 <
-The prefix "RD" in the labels means "cursor down".
+You may close the window opened by `:RMapsDesc` by simply pressing `q`. The
+prefix "RD" in the labels means "cursor down".
```
PMassicotte, line 2148: Do you want to mean that in all the send commands? Should it be placed somewhere else that is better related to the sending commands.

jalvesaq:
> In your `example.Rnw`, at the
> 
> ```r
> # <LocalLeader>fa. Send the functions `plus_one` and `plus_two`.
> ```
> 
> I am getting `Not in an R file` when I try to execute it.

It was my mistake to have included the `<LocalLeader>fa` command in the example. But I will improve the message, making it more generic.

lua/r/send.lua
```diff
@@ -105,15 +105,16 @@ local function get_code_to_send(txt, row)
...
     local node = vim.treesitter.get_node({
         bufnr = 0,
         pos = { row - 1, col - 1 },
-        lang = n_lang,
         -- Required for quarto/rmd/rnoweb; harmless for r.
         ignore_injections = false,
     })
 
+    -- This is a strange fix for https://github.com/R-nvim/R.nvim/issues/298
+    if node then vim.schedule(function() end) end
```
PMassicotte, line 116: I do not understand this code, but apparently it is strange enough if fixes the issues :)

PMassicotte (commented):
Looks good, just a few comments.

doc/R.nvim.txt
jalvesaq: Yes, this is unrelated to the issue being fixed. It's a minor change that I was waiting to commit because I don't like to commit minor changes in the documentation. I prefer to accumulate some of them.

doc/R.nvim.txt
PMassicotte: ah ok from the CLI installation?

lua/r/send.lua
jalvesaq: I guess it fixes a Neovim bug. Compared to Vim, Neovim seems to be much more optimized to avoid unnecessary computations of changes in the user interface while commands are running. It only updates the user interface when the execution of commands finish. My guess is that something needs to be updated immediately and the call to `vim.schedule()` forces the necessary update. It's just a guess.

PMassicotte:
In your example

```r
plus_one <- function(a) {
    # Cursor on `b`. Send the line.
    b <- a + 1
    # <LocalLeader>fc here. Send the function `plus_one`.
    nested_fun <- function(x) {
    # <LocalLeader>fc here. Send the function `plus_one`.
        y <- x + 1
        return(invisible(y))
    }
    b
}
```

If I place my cursor on `plus_one` or `nested_fun` and press enter, it sends the function correctly. Why the `fc` is not working? I might have missed something.

doc/R.nvim.txt
jalvesaq: Yes.

lua/r/send.lua
PMassicotte: I see. LGTM then.

doc/R.nvim.txt
jalvesaq: Yes, the prefix "R" is in all labels, and the letter "D" in "RD" means "Down" and is in many sending commands.  However, users will only need to know this if they are willing to customize the key bindings. So, I think section _8. Custom key bindings_ is the proper place for this information.

jalvesaq:
> If I place my cursor on plus_one or nested_fun and press enter, it sends the function correctly. Why the fc is not working? I might have missed something.

It's because this chunk of R code is actually an R script that I use to check if the use of tree-sitter to send code to R is still working since tree-sitter is frequently changing and breaking our code in `R.nvim`. I should have cleaned up the code and deleted what is supposed to work only in R scripts. From my understanding, the algorithm to send functions is: (1) parse all tree-sitter nodes of the R buffer and get all functions; (2) send the current function according to the cursor position or all functions (depending on the command being `\fc` or `\fa`). To make the code work on Rmd, Quarto, and Rnoweb, the algorithm would have to search for functions in all chunks of R code, but this is not implemented yet.

PMassicotte:
> > If I place my cursor on plus_one or nested_fun and press enter, it sends the function correctly. Why the fc is not working? I might have missed something.
> 
> It's because this chunk of R code is actually an R script that I use to check if the use of tree-sitter to send code to R is still working since tree-sitter is frequently changing and breaking our code in `R.nvim`. I should have cleaned up the code and deleted what is supposed to work only in R scripts. From my understanding, the algorithm to send functions is: (1) parse all tree-sitter nodes of the R buffer and get all functions; (2) send the current function according to the cursor position or all functions (depending on the command being `\fc` or `\fa`). To make the code work on Rmd, Quarto, and Rnoweb, the algorithm would have to search for functions in all chunks of R code, but this is not implemented yet.


If I comment out 

```lua
    -- if vim.bo[bufnr].filetype ~= "r" then
    --     inform("Not yet supported in '" .. vim.bo[bufnr].filetype .. "' files.")
    --     return
    -- end
```

The fa command works.

Peek 2024-12-29 15-49

jalvesaq:
> If I comment out [...] The fa command works in the example.Rnw document

I confirm that it works on my side too! This is great!

jalvesaq:
It doesn't work in Rmd or Quarto, though. So, we have to bring back the previous informational message.

PMassicotte:
> It doesn't work in Rmd or Quarto, though. So, we have to bring back the previous informational message.

It works, but strangely. Let me try to explain below.

jalvesaq:
It doesn't seem to work on `rhelp` (extension `.Rd`) either.

PMassicotte:
1. Create a `file.qmd`

```{r}
#| echo: false
f1 <- function() {
  1 + 1
}

```

2. Close nvim.
3. Open this file in nvim
4. Place cursor on `f1`, press enter. It will send the current function and move down.
6. Place cursor on `f1` again and it works also.
7. Run `localleader fc`, nothing happens.
8. Replace cursor on `f1` and enter again, **it is no longer working**.

jalvesaq:
> Replace cursor on f1 and enter again, it is no longer working.
> It looks like that the buffer type (filetype) is beeing changed somewhere when we try to run fX commands.

I can replicate the same buggy behavior. If we move the cursor, R.nvim recovers the ability to send lines. Maybe, we could try the `vim.schedule()` here too...

PMassicotte:
> > Replace cursor on f1 and enter again, it is no longer working.
> > It looks like that the buffer type (filetype) is beeing changed somewhere when we try to run fX commands.
> 
> I can replicate the same buggy behavior. If we move the cursor, R.nvim recovers the ability to send lines. Maybe, we could try the `vim.schedule()` here too...

The message I am getting when trying to send line after using fX, is `Not inside R or Python code chunk [within markdown]`

jalvesaq:
I just have to move one character above or ahead to bring it to normal behavior.

The `vim.schedule(function() end)` doesn't change anything in this case.

PMassicotte:
Maybe we can merge this PR and open another for the sending behaviour in rmd and qmd files.

jalvesaq:
> Maybe we can merge this PR and open another for the sending behaviour in rmd and qmd files.

I agree, but let's restore the previous informational message before merging it...

jalvesaq:
I suggest:

```lua
    if vim.bo[bufnr].filetype ~= "r" and vim.bo[bufnr].filetype ~= "rnoweb" then
        inform("Not yet supported in '" .. vim.bo[bufnr].filetype .. "' files.")
        return
    end
```

PMassicotte:
Sounds good, Then another issue to fix the `fX` commands.

PMassicotte:
Good job, thanks for the changes.

PR #304, opened 2025-01-08
Fix LSP warnings

jalvesaq:
- Avoid private fields in config class.
- Add missing fields to config class.
- Fix argument types.
- Remove vimdoc links from config.lua.

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -1354,7 +1354,7 @@ example, if it is called by a custom script), you should set the value of
 Set this option in your config if you want to pass command line arguments to
 R at the startup. The value of this variable must be a `string[]`. Example:
 >lua
-   R_args = {"--no-save", "--quiet"}
+   R_args = { "--no-save", "--quiet" }
 <
 On Linux, there is no default value for |R_args|. On Windows, the default
 value is `{"--sdi"}`, but you may change it to `{"--mdi"}` if you do not like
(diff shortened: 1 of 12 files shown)


jalvesaq:
This pull request should not fix or introduce any bugs. It just fixes some LSP warnings that appeared after the Lua class `RConfigUserOpts` definition in `lua/r/config.lua`.

PMassicotte:
I checking it as soon as I am back at office tomorrow morning. But looks great at first! Tx.

jalvesaq:
Maybe we should use only strings for at least some of the options that currently accept more than one type. For example, `external_term` can be boolean or string. We could replace `false` with `""` and `true` with `"default"`.

By the way, I just noted that the `RConfigUserOpts` class registers it as boolean only.

jalvesaq:
For the option `rmdchunk` we could replace `0`, `1`, and `2` with `""`, a single backtick, and `"both"`.

jalvesaq:
> By the way, I just noted that the RConfigUserOpts class registers it as boolean only.

I fixed this and other warnings that became visible when I opened all Lua files at once:

```sh
nvim lua/r/*.lua -p
```

PMassicotte:
This LGTO. I tested it and works fine.

PMassicotte:
> Maybe we should use only strings for at least some of the options that currently accept more than one type. For example, `external_term` can be boolean or string. We could replace `false` with `""` and `true` with `"default"`.
> 
> By the way, I just noted that the `RConfigUserOpts` class registers it as boolean only.

That would be good. I wonder if one can use some kind of schema to define what property values are possible for a given configuration.  Something like this but for lua.

https://json-schema.org/learn/getting-started-step-by-step#create-a-schema-definition

PMassicotte:
> > Maybe we should use only strings for at least some of the options that currently accept more than one type. For example, `external_term` can be boolean or string. We could replace `false` with `""` and `true` with `"default"`.
> > By the way, I just noted that the `RConfigUserOpts` class registers it as boolean only.
> 
> That would be good. I wonder if one can use some kind of schema to define what property values are possible for a given configuration. Something like this but for lua.
> 
> https://json-schema.org/learn/getting-started-step-by-step#create-a-schema-definition

I do not think such thing exists for lua.

PMassicotte:
> We could replace `false` with `""` and `true` with `"default"`.

I agree.

jalvesaq:
Do you mean this?

image

@wurli introduced it with #278. I didn't make it work in my Lazy config, but it works in the R.nvim Lua code.

PMassicotte:
ah true, this is the same on my 

> Do you mean this?
> 
> image
> 
> @wurli introduced it with #278. I didn't make it work in my Lazy config, but it works in the R.nvim Lua code.

ah yeah, I have the same thing here.

PMassicotte:
I we can merge this?

jalvesaq:
Yes, I can change the options `external_term` and `rmdchunk` on another pull request.

PR #305, opened 2025-01-09
Only strings for external_term and rmdchunk

jalvesaq:


jalvesaq:
In this pull request, "string" becomes the only valid type of `external_term` and `rmdchunk` values. This simplifies the code and the documentation.

doc/R.nvim.txt
```diff
@@ -924,10 +924,11 @@ Note: The options of this section are ignored on macOS, where the command
 `open` is called to run the default application used to run shell scripts.
 
 R will run in Neovim's built-in terminal emulator, unless you set the value of
-`external_term`, which can be a boolean value, the name of the terminal
-emulator or the complete command to run the terminal emulator. On Windows,
-only boolean values are valid; if `true` R.nvim will run `Rgui.exe`, and not
-really a terminal emulator.
+`external_term`, which can be `""` (use Neovim's built-in terminal),
```
PMassicotte, line 927: Should not be default? or both "default" and "" are valid for the same thing?

jalvesaq: Do you have any suggestion of how to improve the documentation wording?

PMassicotte: I think this is good. I wrote this comment without reading further. But I think you got it clear.

doc/R.nvim.txt
```diff
@@ -924,10 +924,11 @@ Note: The options of this section are ignored on macOS, where the command
...
+`"default"` (automatically choose a pre-defined terminal), the name of the
+terminal emulator or the complete command to run the terminal emulator. On
+Windows, only boolean values are valid; if `true` R.nvim will run `Rgui.exe`,
+and not really a terminal emulator.
```
PMassicotte, line 931:
```suggestion
terminal emulator or the complete command to run the terminal emulator. 
```

I guess we can remove this since it only accept strings now.

jalvesaq: I'll fix it...

jalvesaq: I changed the text. I think it's still relevant to say that on Windows we don't run R in an external terminal emulator.

PMassicotte:
Thank you for this. This simplify the configuration of `R.nvim`.

jalvesaq:
Thank you for reviewing it so carefully!

PR #318, opened 2025-02-05
Fix quarto render of document with params field

### LINKED ISSUE #315: Unable to knit parameterised Quarto document when parameters already loaded into environment.
Today I have attempted to knit a quarto document after setting parameters and the following error is returned in the R term.

I am trying to render the file using Send to R : quarto quarto_preview.

```
Error: params object already exists in knit environment so can't be overwritten by render params
Execution halted
Error in run_serve_daemon("preview", file, NULL, args, render, port, host,
  :
  Error starting quarto

```
I am running the latest version of R.nvim on Main

Possibly unrelated I am experiencing some C stack usage errors while working interactively.

jalvesaq:
Close #315
Also run stylua

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -1744,7 +1744,7 @@ will consider that the browser was closed and will to open it again.
 For Quarto documents, the available key bindings are <LocalLeader>qr (render
 document), <LocalLeader>qp (preview document), and <LocalLeader>qs (stop
 previewing). The command <LocalLeader>kr achieves basically the same result of
-<LocalLeader>kr. The difference is that the generated document is open after
+<LocalLeader>qr. The difference is that the generated document is open after
 <LocalLeader>kr, and <LocalLeader>qr allows Quarto specific
 options (see |quarto_render_args|).
 
--- a/lua/r/config.lua
+++ b/lua/r/config.lua
@@ -862,10 +862,7 @@ local do_common_global = function()
     config.source_write = config.tmpdir .. "/Rsource-" .. vim.fn.getpid()
 
     -- Default values of some variables
-    if
-        config.RStudio_cmd ~= ""
-        or (config.is_windows and config.external_term ~= "")
-    then
+    if config.RStudio_cmd ~= "" or (config.is_windows and config.external_term ~= "") then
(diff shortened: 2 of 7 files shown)


PMassicotte:
About to go home. Will check it as soon as I am back to office tomorrow morning. But thank you, looks fine from a quick look.

jalvesaq:
A simple Quarto document to replicate the reported bug is:

````rmd
---
params:
    aa: 1
---

Text here.

```{r}
print(params$aa)
```
````

Then, `<LocalLeader>qr`.

I changed the value of the environment variable `R_DEFAULT_PACKAGES` in the running R instance, but, as far as I remember, we never run another R instance from the R Console. So, we don't need the string ",nvimcom" included in the environment variable.

File: doc/R.nvim.txt [hunk 1]
```diff
@@ -1744,7 +1744,7 @@ will consider that the browser was closed and will to open it again.
 For Quarto documents, the available key bindings are <LocalLeader>qr (render
 document), <LocalLeader>qp (preview document), and <LocalLeader>qs (stop
 previewing). The command <LocalLeader>kr achieves basically the same result of
-<LocalLeader>kr. The difference is that the generated document is open after
+<LocalLeader>qr. The difference is that the generated document is open after
```
PMassicotte, line 1747: For me `kr` works but does not open the generated html file from this qmd document

````
---
params:
  aa: 1
---

Text here.

```{r}
print(params$aa)
```
````

PMassicotte:
Tested it with few parametrized documents and noticed no issues. I just have this small comment above with using `<localleader>kr`. I understand it  should open the document once it has been rendered?

doc/R.nvim.txt
jalvesaq: On my side, it sends the `nvim.interlace.rmd()` command to R and opens the HTML document in Firefox. What is the value of the "browser" option in your R installation? Does `browseURL("doc_name.html")` works?

jalvesaq:
> I understand it should open the document once it has been rendered?

Yes. It also depends on the value of `open_html` option.

PMassicotte:
Ha true that. I will merge, good job on this!

PR #324, opened 2025-02-10
Limit search of completion data to .GlobalEnv

jalvesaq:


jalvesaq:
This pull request fixes a very specific completion bug: an error that happens only if completing a function internal to `nvimcom` (not existing in `.GlobalEnv`). For example, if **R is running**, the cursor is at line 298, column 50, of `nvimcom/R/bol.R`, and the Space bar is pressed to trigger the completion.

I checked the `md5sum` of completion data files generated from this branch with the ones generated from the main branch and they are the same which means that no bug is being introduced in the completion data.

PMassicotte:
Good catch. Curious to know how you foud that :)

jalvesaq:
I saw it when removing LSP warnings from nvimcom's R code.

PR #325, opened 2025-02-11
New options replacing rmdchunk and rnowebchunk

### LINKED ISSUE #323: backtick not inserted in correct position in inline r chunk
In quarto and rmd files, when typing the `rmdchunk` option to add an inline code chunk, it adds a single backtick after the subsequent character.

If `rmdchunk` is set to `"both"` (the default), then everything behaves as normal. But as soon as it is changed to something else (such as a single `"'"`, then the `'` inserts a backtick, again in the wrong place. (The `rmdchunk` option insets a full code chunk as normal, when at the start of an empty line.)

Even with `rmdchunk` set to `"both"`, using other plugins (including which [render-markdown.nvim](https://github.com/MeanderingProgrammer/render-markdown.nvim) doesn't set any insert-mode bindings) changes the behaviour of the backtick when adding an inline code chunk.

Using the minimal config below, typing a backtick in insert from: `te|st` (with `|` representing the cursor) should result in the backtick being placed after the `e` but instead adds the backtick after the `s` and moves the cursor after the `s` and before the backtick. So rather than:

```
Te`|st 
```
or 
```
Te`r `st
```

The result is 
```
Tes|`t
```

At the start of a (non-empty) line, the backtick is inserted at the start of the line correctly, but then the cursor is placed to the left of it (rather than the right).

```{lua}
local lazypath = vim.fn.stdpath("data") .. "/lazy/lazy.nvim"
if not (vim.uv or vim.loop).fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "https://github.com/folke/lazy.nvim.git",
    "--branch=stable", -- latest stable release
    lazypath,
  })
end
vim.opt.rtp:prepend(lazypath)

require("lazy").setup({
  {
    "R-nvim/R.nvim",
    dependencies = { "nvim-treesitter/nvim-treesitter" },
    opts = {
      rmdchunk = "`",
    },
  },
})
```

jalvesaq:
Close #323

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -590,7 +590,7 @@ on your system.
 5.3. Object Browser and auto-completion limitations
 
 New lines in list names (which includes column names in data frames and
-matrixes) are replaced with blank spaces in the data used by `cmp-r` and the
+matrices) are replaced with blank spaces in the data used by `cmp-r` and the
 Object Browser.
 
 Objects from environments accessed with the `browser()` function are not
@@ -702,7 +702,7 @@ The compilation of `nvimcom` during R startup will fail on Windows if your
 Send commands to Radian doesn't work well
 
 If you use `radian` as |R_app|, multiple lines will not be correctly
-interpreted because `bracketed_paste` is not recogenized by terminals on
+interpreted because `bracketed_paste` is not recognized by terminals on
 Windows.
 
 
@@ -727,8 +727,8 @@ You may close the window opened by `:RConfigShow` by simply pressing `q`.
 |assignment_keymap|   Choose what to convert into ' <- '
 |pipe_keymap|         Choose what to convert into ' |> ' (or ' %>% ')
 |pipe_version|        Either "native"/"|>" or "magrittr"/"%>%"
-|rnowebchunk|         Convert '<' into '<<>>=\n@' in Rnoweb files
-|rmdchunk|            Convert grave accent chunk into delimiters
+|rnw_chunk_keymap|    Choose what to convert to R chunk delimiters
+|rmd_chunk_keymap|    Choose what to convert to R chunk delimiters
 |rproj_prioritise|    Use .Rproj file to override options
 |objbr_place|         Placement of Object Browser
 |objbr_w|             Initial width of Object Browser window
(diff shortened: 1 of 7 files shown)


jalvesaq:
@rg-ch was the only user who reported bugs on Rnoweb support and may want to comment on this pull request since we are changing the option `rnowebchunk`. If not, I believe this pull request is ready to be merged.

jalvesaq:
Note: I fixed some unrelated typos, including one in `nvimcom.c`.

doc/R.nvim.txt
```diff
@@ -727,8 +727,8 @@ You may close the window opened by `:RConfigShow` by simply pressing `q`.
 |assignment_keymap|   Choose what to convert into ' <- '
 |pipe_keymap|         Choose what to convert into ' |> ' (or ' %>% ')
 |pipe_version|        Either "native"/"|>" or "magrittr"/"%>%"
-|rnowebchunk|         Convert '<' into '<<>>=\n@' in Rnoweb files
-|rmdchunk|            Convert grave accent chunk into delimiters
+|rnw_chunk_keymap|    Choose what to convert to R chunk delimiters
+|rmd_chunk_keymap|    Choose what to convert to R chunk delimiters
```
PMassicotte, line 731: I am not sure to understand what this means. Do you mean something like:

Which key map to insert inline or R code blocks?

lua/r/config.lua
```diff
@@ -286,17 +286,17 @@ local hooks = require("r.hooks")
 ---@field rmd_environment? string
 ---
 ---Controls if and how backticks are replaced with code chunk/inline code
----delimiters when writing R Markdown and Quarto files. Do `:help rmdchunk`
----for more information.
----@field rmdchunk? string
+---delimiters when writing R Markdown and Quarto files.
+---Do `:help rmd_chunk_keymap` for more information.
```
PMassicotte, line 290: I am getting this message:

```
E149: Sorry, no help for rmd_chunk_keymap
```

PMassicotte:
Works fine on my side with both inline or code block. Good job. Just 2 small comments.

lua/r/config.lua
jalvesaq: Maybe lazy.nvim didn't call `:helptags`. It should work if you do it manually:

```
:helptags /path/to/R.nvim/doc
```

PMassicotte: It is working now. Not sure why it was not when I used the PR checkout.

doc/R.nvim.txt
PMassicotte: Here we are talking about which keymap right? Not the character to be used like "`" or something else?

PR #326, opened 2025-02-12
fix(send.lua): prevent cursor from moving beyond the last line by clamping end_row to buffer line count

PMassicotte:
This fix a small issue when using the `<localleader>fd` in the case the function is at the end of the buffer.

For example, this R file with just this code:

```r
f1 <- function(x) {
  x + 1
}
```

With the cursor inside, press `<localleader>fd` and it will crash because we are trying to move the cursor past the end of the buffer.


jalvesaq:
Thanks! I confirm the bug and that its fixed by the pull request!

PR #327, opened 2025-02-12
Replace some options with default commands

jalvesaq:
Remove assignment_keymap, pipe_keymap, rmd_chunk_keymap and rnw_chunk_keymap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,8 +182,8 @@ firm commitment to backwards compatibility.
 
 ### Changes:
 
-- `<M-->` (i.e. `Alt + -`) is now used to insert `<-`. This can be configured
-    using `assignment_keymap`.
+- `<M-->` (i.e. `Alt + -`) is now used to insert `<-`. See the documentation
+  on how to create custom key bindings.
 
 - `R_source` and `after_R_start` have been replaced with more powerful `hook`
     options.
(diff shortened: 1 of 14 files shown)


jalvesaq:
As suggested at #323, this pull request proposes converting three options into conventional key mappings which can be changed in the `on_filetype` hook or disabled with the option `disable_cmds`. I adapted and moved part of the documentation from the section "Options" to the section "Use".

I noted that the `on_filetype` hook can't be run within a `vim.schedule()` command because the maps created in the hook would not be available when the default mappings were created. The function `vim.schedule()` postpones the execution of its argument until Neovim is idle. So, I added a third argument to the `hooks.run()` function.

PMassicotte (commented):
All good on my side, impressive rework. Found no conflicts or breaking changes on my side. 

Just one thing, when typing `:RMapDesc`, is it expected to see missing some keymap?

image

wurli:
Looks great to me! A minor comment: any chance we can use a common terminology for the insert commands? E.g. `RInsertAssign`, `RInsertPipe`, `RnwInsertChunk`, `RmdInsertChunk`?

jalvesaq:
The bug of repeating the mode is fixed.

PMassicotte:
> Looks great to me! A minor comment: any chance we can use a common terminology for the insert commands? E.g. `RInsertAssign`, `RInsertPipe`, `RnwInsertChunk`, `RmdInsertChunk`?

I think that could be a good idea.

jalvesaq:
I missed this comment, but indeed it's a good idea. I make the changes...

PMassicotte:
LGTM! Thanks for the work.

PMassicotte:
Ready to be merged?

PR #331, opened 2025-02-16
New function for health.lua: check_lang()

jalvesaq:


PMassicotte:
Thank you! Works fine on my side.

PR #332, opened 2025-02-16
Remove integration with Rapp on macOS

### LINKED ISSUE #309: Remove support for R.app if not used
Does anyone have this option in the `R.nvim` config?

```lua
applescript = true
```

If nobody uses `R.app`, I will remove the above option from `R.nvim` at the end of February 2025.

jalvesaq:
Close #309


PMassicotte:
Than you!

PR #334, opened 2025-02-19
feat(buffer): add function to create R buffer from RMarkdown or Quart…

PMassicotte:
This PR enables Tree-sitter support for QMD and RMD files. The core addition is the new function:  

https://github.com/R-nvim/R.nvim/blob/403e956d243397fb12681af66b5e775fbc48d9ff/lua/r/buffer.lua#L12-L79

This function creates a temporary R buffer by stripping out non-R code, allowing seamless integration with Tree-sitter tools. Instead of `vim.api.nvim_get_current_buf()`, we can now use `create_r_buffer()`, ensuring Tree-sitter operates on pure R content.  

As a practical example, I refactored the function-sending logic to the console. It now works transparently across different file types:  
https://github.com/R-nvim/R.nvim/blob/403e956d243397fb12681af66b5e775fbc48d9ff/lua/r/send.lua#L784

### DIFF
--- a/lua/r/buffer.lua
+++ b/lua/r/buffer.lua
@@ -1,0 +1,81 @@
+local inform = require("r.log").inform
+local warn = require("r.log").warn
+
+local M = {}
+
+local get_root_node = require("r.utils").get_root_node
+
+--- Creates an R buffer from the current buffer or a specified buffer number.
+--- If the current buffer is already an R file, it returns the current buffer number.
+--- For Quarto, or RMarkdown files, it extracts R code chunks and creates a new buffer.
+--- @return buffer|nil The buffer number of the created R buffer, or nil if creation fails.
+M.create_r_buffer = function()
+    local bufnr = vim.api.nvim_get_current_buf()
+
+    local filetype = vim.bo[bufnr].filetype
+
+    if filetype == "r" then return bufnr end
+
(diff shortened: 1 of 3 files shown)


jalvesaq:
In a quick test, the behavior of `<LocalLeader>fc` and `<LocalLeader>fa` didn't change, but the cursor jumps to the wrong position after `<LocalLeader>fd` on the "main" branch, and to the correct position on the "rbuff" branch. Are there other advantages that I'm missing?

Anyway, it's very tricky to work with treesitter and I'm learning with your work in `R.nvim`. Good job!

PMassicotte:
> Are there other advantages that I'm missing?

Anything we can do on line/node selection will be much easier. We could also refactor, when we have time, all the code that select the chunks (send all chunks, send chunks above, etc) without having to use regex to capture them.  One thing I use a lot is the chain breaker when I am working on code:

https://github.com/R-nvim/R.nvim/blob/d3083515cdac6fd12ab67b7530abb2c3504251e9/lua/r/send.lua#L707-L739

This could be refactored to use pure TS and be more rebust.

jalvesaq:
> This could be refactored to use pure TS and be more rebust.

This would be great! I'm merging this pull request for now...

PR #336, opened 2025-02-19
docs(R.nvim.txt): remove outdated example of using Lua function for `open_app` configuration

PMassicotte:
Remove the example using a lua function for the `open_app` parameter, since we are now accepting only string.

https://github.com/R-nvim/R.nvim/blob/48c0cafdad3baddc350dea9d68f3bc3d1db3065a/lua/r/config.lua#L111

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -1957,18 +1957,7 @@ Calc:
    open_app = "localc"
    open_app = "c:/Program Files (x86)/LibreOffice 4/program/scalc.exe"
 <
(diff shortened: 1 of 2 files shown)


jalvesaq:
We still have this in `lua/r/edit.lua`:

```lua
    if type(open_app) == "function" then
        open_app(tsvnm)
        return
    end
```

Shouldn't we delete these lines too?

PMassicotte:
Good catch, I have removed it and also the type since `open_app` is always a string.

jalvesaq:
Thanks!

PR #340, opened 2025-02-23
Add custom keymaps to :RMapsDesc

### LINKED ISSUE #335: "custom" in RMapsDescs (feature)
Is there a way to display the explicit defined mapping in  `:RMapsDesc` instead of the term "custom", we could keep the same color (violett) in order to distinguish defaults from custom.
In my case, see `RDSendline` or `RSendSelection` below.
The more you define custom mappings the more `:RMapsDesc` risks to become useless.
I can imagine that's not so easy because user may define the same mapping in different places or ways, isn't it ?
Or is there a way to define custom mappings in order to make them display explicitly in  `:RMapsDesc` 
 
It's not a bug. I don't know where to ask for the improving feature.

Image

Thanks in advance for your great work  !

jalvesaq:
Close #335

### DIFF
--- a/lua/r/maps.lua
+++ b/lua/r/maps.lua
@@ -384,9 +384,17 @@ M.show_map_desc = function()
             )
             table.insert(map_key_desc, { string.format("  %-04s", v[2]), "Type" })
             local keymap = v[3] or " "
+            local syngrp = keymap == "disabled" and "Comment" or "Special"
+            local modes = vim.split(v[2], "")
+            for _, m in pairs(modes) do
+                if get_map_to(v[1], m) ~= "" then
+                    syngrp = "String"
+                    break
+                end
+            end
             table.insert(map_key_desc, {
                 string.format("%-0" .. kw .. "s", keymap),
-                (keymap == "custom" or keymap == "disabled") and "Comment" or "Special",
+                syngrp,
             })
             table.insert(map_key_desc, { v[4] .. "\n" })
         end
--- a/lua/r/maps.lua
+++ b/lua/r/maps.lua
@@ -84,14 +84,45 @@ local map_desc = {
     ROBToggle =     { m = "", k = "", c = "Object_Browser", d = "Toggle the Object Browser" },
 }
 
-local map_it = function(plug, mode, plg, tgt, cmd, opts)
+local map_list
+
+local fill_map_list = function()
+    local plg = ""
+    local map = ""
+    map_list = { n = {}, v = {}, i = {}}
+    for _, mode in pairs({"n", "i", "v"}) do
+        local txt = vim.fn.execute(mode .. "map")
+        local lines = vim.split(txt, "\n")
+        for _, v in pairs(lines) do
+            if v:find("<Plug>R") and not v:find("^" .. mode .. "  <Plug>R") then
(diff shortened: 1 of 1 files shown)


PMassicotte:
One thing I noticed when typed `:RMapsDesc`, 

```
 RSendAboveLines       n   ;su                   Send to R all lines above the current one
```

This is the only one using `;` instead of `<LodalLeader>`.

PMassicotte:
I understand I should see any custom mapping that use `<Plug>Rxxx` right?

I have this,

```lua
hook = {
    on_filetype = function()
        local map_opts = { noremap = true, silent = true, buffer = 0 }

        -- Function to set key mappings
        local function set_keymap(mode, lhs, rhs, opts)
            vim.keymap.set(mode, lhs, rhs, opts or map_opts)
        end

        -- Key mappings
        set_keymap("n", "<M-S-CR>", "<Plug>RSendChain", { silent = true })
        set_keymap("i", "<M-S-CR>", "<C-o><Plug>RSendChain", { silent = true })
    end,
}
```

However, I do not see them in the `RMapsDesc` list, it is still showing it as `<LocalLeader>sc`.

image

jalvesaq:
`RSendChain` is mapped only in Normal and Visual modes in `lua/r/maps.lua`:

```lua
    create_maps("nv", "RSendChain",      "sc", "<Cmd>lua require('r.send').chain()")
```

You are creating keybindings for Normal and Insert modes, but the Visual mode keybinding prevails in the `:RMapsDesc` output.

Perhaps, it doesn't make sense to create a keybinding to `RSendChain` in Visual mode, and we should create the keybinding in Insert mode. That is, the above code would become:

```lua
    create_maps("ni", "RSendChain",      "sc", "<Cmd>lua require('r.send').chain()")
```

I guess we should review the argument  `mode` when we call the function `create_maps`. The function is called 100 times:

```
grep create_maps lua/r/maps.lua | wc
```

PMassicotte:
> ```
> 
> ```
> 
> 
>     
>   
> 
> I don't know why it's `;su` on your side. What's the corresponding output of `:nmap`?

I have two occurances of `rsendabove` in `:nmap`:

```
n  <Plug>RSendAboveLines *@<Cmd>lua require('r.send').above_lines()<CR>
                 Send to R all lines above the current one
```

and 

```
n  ;su         * <Plug>RSendAboveLines
                 Send Above Lines
```
If you are using Telescope, you can also type `:Telescope keymaps`  to get a fuzzy search on your mappings

image

PMassicotte:
> Perhaps, it doesn't make sense to create a keybinding to `RSendChain` in Visual mode, and we should create the keybinding in Insert mode. That is, the above code would become:

I have removed the mapping of the insert mode, but I still see the `<localleader>sc` in `RMapsDesc`. The only custom mapping I have to sending chain is 

```lua
set_keymap("n", "<M-S-CR>", "<Plug>RSendChain", { silent = true })
```

PMassicotte:
For some reasons, it is working now with your good idea to use another color.

image

Ready to be merged? Looks like it is working fine.

jalvesaq:
Yes, it's ready.

PMassicotte:
Thank you!

PR #342, opened 2025-02-23
refactor(numbers.lua, quarto.lua): generalize code chunk retrieval function

### LINKED ISSUE #338: Parameterize get_r_chunks_from_quarto
- Rename to `get_chunks_from_quarto =function(root, bufnr, cursor_pos, lang)`

> Good call. I will parameterize it so we can specify pass the chunk language:
> 
> https://github.com/R-nvim/R.nvim/blob/6ff074d59431bb051e3d786bd35a081916065ad3/lua/r/quarto.lua#L29-L38 

 _Originally posted by @PMassicotte in [#337](https://github.com/R-nvim/R.nvim/issues/337#issuecomment-2672609154)_

PMassicotte:
- Let us use the function to retrieve code chunks in a rmd/qmd document, so we can only use TS to capture any code chunk.

- If/when we refactor the send chunks functions, we could use it for both python and R. 

closes #338

### DIFF
--- a/lua/r/format/numbers.lua
+++ b/lua/r/format/numbers.lua
@@ -1,6 +1,5 @@
 local inform = require("r.log").inform
 local get_code_chunk = require("r.quarto").get_code_chunk
-local get_root_node = require("r.utils").get_root_node
 
 local M = {}
 
@@ -64,11 +63,13 @@ M.formatnum = function(bufnr)
         return
     end
 
-    local root_node = get_root_node(bufnr)
-    if not root_node then return end
-
     if filetype == "quarto" or filetype == "rmd" then
-        local r_chunks_content = get_code_chunk(root_node, bufnr, nil, "r")
+        local r_chunks_content = get_code_chunk(bufnr, "r")
+
+        if not r_chunks_content then
+            error("Failed to extract code chunks.")
+            return
+        end
 
         for _, r_chunk in ipairs(r_chunks_content) do
             find_and_replace_float(r_chunk.content, bufnr, r_chunk.start_row, 0)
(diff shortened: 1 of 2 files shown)


jalvesaq:
Currently, the code looks like this:

image

Using `:Neogen` command to insert the function documentation template, I get this:

image

The template inserted by the `:Neogen` command uses three dashes instead of two and has no space between the dashes and the argument name. This format gets a different highlight that helps to spot mistakes. More specifically, the type of each argument was missing.

PMassicotte:
Let me install it. Tx for the tips

jalvesaq:
Since `cursor_pos` is optional, perhaps, it should be the last argument. And because it's only the row number and not the complete cursor position, its name could possibly be `row`.

The argument `root` can be derived from `bufnr`. So, perhaps the code to get the root node could be moved to the `get_code_chunk` function (unless you foresee cases where the root node will not be the root of the buffer).

jalvesaq:
The argument `code_fence_language` could be renamed as `language`, or abbreviated to `lang` because we already have "code" and "chunk" in the name of the function.

PMassicotte:
Does neogen also generate the description?

Here it just generated 

```
---@param root 
---@param bufnr 
---@param cursor_pos 
---@param code_fence_language 
---@return 
```

jalvesaq:
> Here it just generated

Yes, just that. I copied the description from the existing comments. Then:
 - The missing type became obvious because the words "The" and "A" were highlighted with the Type color group.
 - The missing last argument was included.

PMassicotte:
I should not have missed any of your suggestions. Let me know.

jalvesaq:
It seems that for treesitter Rmd and Quarto are the same file type. Hence, the description of `get_code_chunk()` could be:

```
--- Helper function to get code block from Rmd or Quarto document
```

jalvesaq:
Merging now... Thanks!

PR #343, opened 2025-02-24
refactor(numbers.lua): simplify number formatting

PMassicotte:
No longer need to care about the filetype to format numbers. It is using the new `create_r_buffer()` that allows to operate on R, qmd and rmd file seamlessly.


jalvesaq:
Good job! Same result with less code!

PMassicotte:
Tx!

PR #347, opened 2025-02-25
Minor refactor

### LINKED ISSUE #346: Can't restart an R session once it is closed
I use an external terminal to run the R instance. There are times when I start R (`<leader>rf`) and for whatever reason, it opens an R instance, but doesn't connect to it (I can't send a line for instance). Usually, I would then close it and restart it. Recently, once the R instance has been closed, it cannot be restarted with `<leader>rf`. The only way then is to quit the script, re-open, and open the R instance again. Am I missing something?

jalvesaq:
- Avoid repetion of module name in function name.
- Move Rgui functions to rgui.lua
- Also, deep copy table (Fix #346)

### DIFF
--- a/lua/r/external_term.lua
+++ b/lua/r/external_term.lua
@@ -114,7 +114,7 @@ end
 
 local M = {}
 
-M.start_extern_term = function()
+M.start = function()
     if config.config_tmux then
         -- Create a custom tmux.conf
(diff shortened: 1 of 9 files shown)


PMassicotte:
Excellent, I confirm that I am able to re-open the terminal once close. Good job.

PR #355, opened 2025-03-03
New get_lang()

### LINKED ISSUE #353: incompatibility with md-headers, a plugin to navigate through the headings of .md .rmd.
Not an issue, but an incompatibility between packages. Must I maybe put it on "Discussions"?

I installed the https://github.com/AntonVanAssche/md-headers.nvim/wiki/Installation and loved the idea to easily navigate through the headers. Maybe in the future can I fork it and try to implement a chunk navigation, like in RStudio.  But for now, my knowledges in Lua is very limited.
Well, after installing the plugin (I'm using Lazyvim) and enabling more formats `ft = { "markdown", "rmd", "Rmd", "qmd" }`, everything works fine until I call any function of the package, like `MDHeaders` or `MDHeadersTelescope`. After that, I cannot send code to terminal "not inside R or Python Chunk" and when I try to comment, it uses HTML comment `<!--`  everywhere, instead of `#`.  Has anyone any clue on how to solve this problem? Or knows anybody an alternative, similar plugin? For me, the best solution would be to easily navigate through the names of the chunks. 

Thanks!

jalvesaq:
Close #353

### DIFF
--- a/lua/r/quarto.lua
+++ b/lua/r/quarto.lua
@@ -62,4 +62,41 @@ M.get_code_chunk = function(bufnr, lang, row)
     return r_contents
 end
 
+--- Helper function to get code block info from Rmd or Quarto document
+---@return table|nil
+M.get_ts_chunks = function()
+    local root = require("r.utils").get_root_node()
+
+    if not root then return nil end
+
+    local query = vim.treesitter.query.parse(
+        "markdown",
+        [[
+           (fenced_code_block) @fcb
+        ]]
+    )
+
+    local bufnr = vim.api.nvim_get_current_buf()
+    local chunks = {}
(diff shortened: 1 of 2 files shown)


jalvesaq:
This is an alternative way of checking if the cursor is within a code block that seems to be more robust than the current code.

jalvesaq:
Yes, your pull request also fixes the issue and is more complete.

jalvesaq:
Actually neither of the two pull requests seems to fix the issue... I forgot to run `:MDHeaders`.

PMassicotte:
> Actually neither of the two pull requests seems to fix the issue... I forgot to run `:MDHeaders`.

Yeah, just found that. Tomorrow I will take time to investigate this.

PR #357, opened 2025-03-06
Use quarto module to get the lang of the current chunk

PMassicotte:
This PR uses the new quarto module to determine language type in the `get_lang()` function.

### DIFF
--- a/lua/r/send.lua
+++ b/lua/r/send.lua
@@ -517,6 +517,7 @@ end
 M.line = function(m)
     local lnum = vim.api.nvim_win_get_cursor(0)[1]
     local line = vim.fn.getline(lnum)
+
     local lang = get_lang()
     if lang == "chunk_child" then
         if type(m) == "boolean" and m then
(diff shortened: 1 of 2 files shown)


jalvesaq:
Thanks!

Comparing this code

```lua
    local lang_map = {
        rnoweb = get_rnw_lang,
        rhelp = get_rhelp_lang,
    }

    if lang_map[filetype] then return lang_map[filetype]() end
```

with the previous one

```lua
    -- Treesitter for rnoweb always return "latex" or "rnoweb"
    if vim.bo.filetype == "rnoweb" then return get_rnw_lang() end
    -- No treesitter parser for rhelp
    if vim.bo.filetype == "rhelp" then return get_rhelp_lang() end
```

I admit that your version is more elegant because it calls a single function for two cases, but the previous code was simpler to understand and computationally less expensive.

PMassicotte:
Ok I will change it back :)

jalvesaq:
If the cursor is over the line:

````
```{r}
````

The line should not be sent to R. On the main branch, we get an informational message:

```
Not inside R or Python code chunk [within chunk_header]
```

With the pull request, the line is sent to R and we get this error:

```
Error: attempt to use zero-length variable name
```

PMassicotte:
Good catch. Will fix.

PR #366, opened 2025-03-14
Improve control of chunk highlighting (close #365)

### LINKED ISSUE #365: Code syntax highlighting missing
Hi I have the main branch turned on on both Windows and mac. In both instances with the most recent update loaded via Lazy code highlighting is dropped.

The quarto chunk highlight block is visible, but the contents of chunks are plaintext. Code still runs as expected. The issue persists in R files. In the same session YAML files are highlighted appropriately.

Reverting to commit f10e7f0871a6854f65c76269846a69c7232146bd (the previous commit on  main) restores syntax highlighting.

jalvesaq:
- Highlight YAML in python code
- New option: yaml_hl
- Wrap the call to query.set with vim.schedule

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -1714,18 +1714,19 @@ config, as in the examples:
 <
 R.nvim changes the background color of code blocks and adds a virtual text
 with the language of the code block to the right side of the header of the
-block. It also highlights comments prefixed with `#|` with the foreground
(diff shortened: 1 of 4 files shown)


PMassicotte:
All good on my side! 

Thank you.

jalvesaq:
Thanks!

PR #370, opened 2025-03-15
Fix get_lang() for cmp_r

jalvesaq:
See https://github.com/R-nvim/cmp-r/pull/13

### DIFF
--- a/lua/r/utils.lua
+++ b/lua/r/utils.lua
@@ -80,25 +80,23 @@ function M.get_lang()
     local filetype = vim.bo.filetype
     if filetype == "" then return "none" end
 
-    -- Handle specific filetypes early
+    -- Handle specific filetypes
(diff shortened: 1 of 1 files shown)


jalvesaq:
Changes:
- Use `vim.treesitter.get_parser()` instead of `require("nvim-treesitter.parsers").get_parser()` because the former returns "yaml", "markdown" and "markdown_inline" whereas the latter returns only "markdown".
- Fix row number when calling `parser:language_for_range()`.

PMassicotte (approved):

PMassicotte:
How did you find that for the parser? Very strange that both does not return the same thing.

Tx

jalvesaq:
The old code called `vim.treesitter.get_parser()`. I don't remember who wrote the old code.

Issue #359, opened 2025-03-07
No R or Python code chunk found at the cursor position


idavydov:
The latest version stopped recognizing certain R chunks in R Markdown.

Steps to reprocude:
1. Navigate to chunk
2. `<localleader>cc`

Behaviour: Message "No R or Python code chunk found at the cursor position"

**not recognized**, not desired behavior
````
```{r, eval=!FALSE}
print("hi")
```
````

**not recognized**, potentially not desired behavior, since these are instructions for `knitr` and it should be still possible to trigger execution manually (`<localleader>cc`)
````
```{r, eval=FALSE}
print("hi")
```
````

Could be potentially related to #353 , but i'm not sure

**Comment by PMassicotte** [comment]
Try with `eval=TRUE`

**Comment by PMassicotte** [comment]
We could potentially run them even if it is set to FALSE.

**Comment by idavydov** [comment]
`eval=TRUE` runs the code successfully

**Comment by idavydov** [comment]
And just to give background, I sometimes have R markdown chunks like this:

````
---
params:
  do_something: true
---
```{r, eval=do_something}
something()
```

```{r, eval=!do_something}
something_else()
```
````

**Comment by PMassicotte** [comment]
Maybe we could just

In the "old" code (https://github.com/R-nvim/R.nvim/commit/508dc92f39c7adf2c6a48fedeac1c073992d363b#diff-d49f74bb08746ae5d64f2873a7321f320961815717bceaca7adbbf0defb2d175L355) I think the `eval=false` chunks were also skipped. @jalvesaq ?

**Comment by idavydov** [comment]
In my opinion the correct behavior should be this:

-  `<localleader>cc` should ignore the `eval` value as it is called expliclity
- `<localleader>ch` could consider `eval` but then it should properly evaluate the expression; if this is doable

**Comment by jalvesaq** [comment]
I'm sorry. I will only be able to look at this on Sunday.

**Comment by jalvesaq** [comment]
Is it fixed after https://github.com/R-nvim/R.nvim/commit/6604d622a6a5c98149c15cbcf089c7672c7fc4f5 ?

**Comment by jalvesaq** [comment]
Currently, with this Quarto document:

````
Initial paragraph.

```{r}
x <- 1
```

Text.

```{r}
#| eval: false
x <- 2
```

More text

```{r}
print(x)
```

````

- If the cursor is over the last (empty) line of the document and we do `<LocalLeader>ch`, the second chunk is skipped. I think this is the ideal behavior.
[...]

Issue #368, opened 2025-03-14
Allow sending the current chunk even if `eval: false`


PMassicotte:
> Currently, with this Quarto document:
> 
> ````
> Initial paragraph.
> 
> ```{r}
> x <- 1
> ```
> 
> Text.
> 
> ```{r}
> #| eval: false
> x <- 2
> ```
> 
> More text
> 
> ```{r}
> print(x)
> ```
> 
> ````
> 
> - If the cursor is over the last (empty) line of the document and we do `<LocalLeader>ch`, the second chunk is skipped. I think this is the ideal behavior.
> - If the cursor is over `x <- 2` and we try to send the line to R, the line is sent. I think this is correct too.
> - If the cursor is over the header of the second chunk, we get the message "No evaluable R or Python code chunk found at the current cursor position". I think this is OK because the user might be sending chunks and jumping to the next in a quick succession, without paying attention to what is being sent to R until reaching a specific chunk. It's similar to `<LocalLeader>ch`, but more manual.
> - If the cursor is over `x <- 2` and we do `<LocalLeader>cc`, we also get the message "No evaluable R or Python code chunk found at the current cursor position". I think this could be different because the user is more thoughtfully trying to send the chunk to R. In my opinion, in this case, we could ignore the value of `eval` and send the chunk to R. 

 _Originally posted by @jalvesaq in [#359](https://github.com/R-nvim/R.nvim/issues/359#issuecomment-2725293067)_

PMassicotte:
The sending logic should be reviewed a bit. I do not think that the filtering should be handled in the `send_*()` functions, but rather upstream.

https://github.com/R-nvim/R.nvim/blob/78a282243ffa546b801a443c9403eb5fc21ec232/lua/r/rmd.lua#L56-L79

jalvesaq:
Maybe

```lua
     local chunks = quarto.get_current_code_chunk(bufnr) 
```

could be

```lua
     local chunks = quarto.get_current_code_chunk(bufnr, eval) 
```

Where `eval` would indicate if `eval: false` is acceptable.

PMassicotte:
Very nice! Idea

```lua
     local chunks = quarto.get_current_code_chunk(bufnr, include_eval) 
```

jalvesaq:
Closing this after https://github.com/R-nvim/R.nvim/pull/372...

PR #372, opened 2025-03-18
fix(quarto.lua): correct logic for evaluating code chunks

PMassicotte:
This pull request includes a smal change to the `lua/r/quarto.lua` file. The change modifies the logic for evaluating code chunks based on their parameters.

* [`lua/r/quarto.lua`](diffhunk://#diff-a628cdf45a9f0894d37ede53c461854baf4f8b048983d06183ea64ae08b33d5aL349-R352): Updated the `M.filter_code_chunks_by_eval` function to correctly handle the `eval` parameter by changing the condition to check if `eval` is not equal to "true" or "TRUE" instead of equal to "true" or "TRUE".

It prevents running the chunk when is explicitly set to false.

See https://github.com/R-nvim/R.nvim/issues/359#issuecomment-2733887878

### DIFF
--- a/lua/r/quarto.lua
+++ b/lua/r/quarto.lua
@@ -346,10 +346,10 @@ M.filter_code_chunks_by_eval = function(chunks)
 
         -- Check for eval in comment_params
         if chunk:get_comment_params() and chunk:get_comment_params().eval then
-            eval = chunk:get_comment_params().eval == "true"
+            eval = chunk:get_comment_params().eval ~= "true"
         -- Check for eval in info_string_params
         elseif chunk:get_info_string_params() and chunk:get_info_string_params().eval then
-            eval = chunk:get_info_string_params().eval == "TRUE"
+            eval = chunk:get_info_string_params().eval ~= "TRUE"
         end
 
         return eval -- Return true if eval is "true"
--- a/lua/r/quarto.lua
+++ b/lua/r/quarto.lua
@@ -346,10 +346,10 @@ M.filter_code_chunks_by_eval = function(chunks)
 
         -- Check for eval in comment_params
         if chunk:get_comment_params() and chunk:get_comment_params().eval then
-            eval = chunk:get_comment_params().eval ~= "true"
+            eval = chunk:get_comment_params().eval ~= "false"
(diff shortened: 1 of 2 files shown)


lua/r/quarto.lua
```diff
@@ -346,10 +346,10 @@ M.filter_code_chunks_by_eval = function(chunks)
 
         -- Check for eval in comment_params
         if chunk:get_comment_params() and chunk:get_comment_params().eval then
-            eval = chunk:get_comment_params().eval == "true"
+            eval = chunk:get_comment_params().eval ~= "true"
```
idavydov, line 349: shouldn't it be `~= "false"`?

jalvesaq:
It doesn't seem to be correct with this Quarto document:

````
---
title: Testing the eval option
---

```{r}
x <- 1
```

Text

```{r}
#| eval: false

x <- x + 1
```

Text

```{r}
x <- x + 1
```

Text

```{r}
print(x)
```

````

- With the cursor on the last line, after `<LocalLeader>ch`, the value of `x` should be 2, not 3.
- With the cursor on the header of the first chunk, sending a "line", should send the chunk and jump to the third chunk.

PMassicotte:
oups, should be ok

jalvesaq:
Now, we have tho opposite problem. When the cursor is over `x <- x + 1` in the `eval: false`  block, `<LocalLeader>cc` should work because the user really wants to send the block.

PMassicotte:
> Now, we have tho opposite problem. When the cursor is over `x <- x + 1` in the `eval: false` block, `<LocalLeader>cc` should work because the user really wants to send the block.

This is still not implemented (ie sending false chunk when inside it)

https://github.com/R-nvim/R.nvim/issues/368

jalvesaq:
With the suggested change, if the cursor is over an eval-false chunk header, the chunk is sent with the send-line command. But this seems acceptable to me because the chunk isn't sent by `<LocalLeader>ch` and is skipped when sending chunks sequentially.

PMassicotte:
yes, I think that could do it. I will add this to this PR.

jalvesaq:
Thanks! Merging it now.

PR #374, opened 2025-03-18
Minor changes

jalvesaq:
- Fix highlighting of YAML comments in development version of Neovim.
- Fix LSP warnings.


jalvesaq:
The current development version of Neovim stopped highlighting YAML comments in code blocks of Quarto documents. The fix was simple: don't call `setup_yaml_hl()` within a `vim.schedule()` command.

The other changes fix LSP warnings caused by the mix of string and boolean types. Now, the argument received by `send.line()` is a string (not a boolean or string).

PR #386, opened 2025-04-05
Fix `after_ob_open` hook

jalvesaq:


jalvesaq:
Currently, the `after_ob_open` hook is scheduled to run when Neovim is idle. The result is that the current buffer is no longer the Object Browser when the hook is executed, and commands that depend on the correct buffer number will not work. For example, if the `ccc.nvim` plugin is installed, objects whose names coincide with color names or codes will be highlighted in the Object Browser with the corresponding color. To replicate the issue:

- Install the [ccc.nvim](https://github.com/uga-rosa/ccc.nvim) plugin.
- Start R, open the Object Browser (`<LocalLeader>ro`), and send the following line to R:

```r
dtfrm <- data.frame(f00 = "red", af0 = "#af0", nocolor = 0)
```

With this pull request, the following configuration will work as expected:

```lua
  hook = {
    after_ob_open = function()
      vim.cmd("CccHighlighterDisable")
    end
  }
```

PMassicotte:
I confirm it works on my side. Thanks!

PMassicotte:
I should use more the hook system you implemented.

jalvesaq:
> I should use more the hook system you implemented.

Then we could improve the documentation (or Wiki) with more examples...

PR #392, opened 2025-05-04
Set Object Browser size effectively

jalvesaq:


jalvesaq:
I noted that Neovim isn't respecting the values of `config.objbr_w` when I was trying to use the `after_ob_open` hook to split the Object Browser window to display the Aerial outline:

```lua
                    after_ob_open = function()
                        vim.cmd("split")
                        local buf1 = vim.api.nvim_win_get_buf(1000)
                        if vim.bo[buf1].filetype == "quarto" then
                            local ob2 = vim.api.nvim_get_current_win()
                            require("aerial").open_in_win(ob2, 1000)
                        end
                    end
```

This pull request postpone the resizing to when Neovim is idle.

PMassicotte (approved):
That is excellent, and good catch also!

jalvesaq:
For anyone reading this thread, the `split` command is at the wrong place in my comment above. The correct function for the `after_ob_hook` is:

```lua
                    after_ob_open = function()
                        local buf1 = vim.api.nvim_win_get_buf(1000)
                        if vim.bo[buf1].filetype == "quarto" then
                            vim.cmd("split")
                            local ob2 = vim.api.nvim_get_current_win()
                            require("aerial").open_in_win(ob2, 1000)
                        end
                    end
```

Issue #241, opened 2024-10-05
Support WezTerm mux splits


JonGretar:
I'm not quite familiar with how R.nvim handles `external_term` but I would love to be able to open R in a WezTerm pane much like in a tmux pane. This should work for darwin much like tmux.

I guess in a Linux machine I could just set the full command but I am not sure why it is not allowed for darwin. I'm sure there are good reasons but I don't know why this would not be possible with WezTerm much like with tmux.

The command to run would be something like `wezterm cli split-pane --bottom --percent 30 -- R`

jalvesaq:
It's possible with Kitty. Please, see the last comments on https://github.com/R-nvim/R.nvim/issues/115. You can try to adapt the solution for WezTerm.

JonGretar:
> It's possible with Kitty. Please, see the last comments on #115. You can try to adapt the solution for WezTerm.

As I understand it. Then on MacOS R.term only allows `true` or some tmux command in external_term. Am I incorrect about this?

JonGretar:
Figured out that I could fix this by lying to R.nvim about it being installed on a mac.

```lua
local options = {
   is_darwin = false,
    external_term = 'wezterm cli split-pane --bottom --percent 30  -- ',
 }
```

JonGretar:
Got this working by lying to R.nvim about being on a mac.

```lua
    local options = {
      is_darwin = false,
      external_term = 'wezterm cli split-pane --bottom --percent 30  -- ',
      R_app = 'radian',
    }
```

jalvesaq:
Thanks for sharing your solution!

The last time that I tried Nvim-R on a Mac was in 2015, and, as far as I remember, OS X always started R in whatever was the default terminal. Did macOS inherit this feature (or limitation)? You, people who have access to a macOS, can decide whether the `is_darwin` config option is still necessary or not.

jalvesaq:
Another option that I don't know if still works on macOS is `applescript = true`.

JonGretar:
> Thanks for sharing your solution!
 > The last time that I tried Nvim-R on a Mac was in 2015, and, as far as I remember, OS X always started R in whatever was the default terminal. Did macOS inherit this feature (or limitation)? 

This seems to be a design choice that R.nvim itself makes. By using the `open` command to launch when external_term is set to true. The `open` command simulates what would happen if double click on the file in the GUI file finder unless it is set to `tmux`. But as I am using a command to interface to the terminal that I am running this does not seem to work so well.

[...]

PR #406, opened 2025-06-29
Remove broken support for rgui.exe

### LINKED ISSUE #308: Communication with Rgui.exe doesn't work
As reported by @oktouerl, `Rgui.exe` as an external "terminal" doesn't work: https://github.com/R-nvim/R.nvim/issues/307#issuecomment-2585240939

jalvesaq:
- Same default `R_args` for Linux and Windows.
- Remove options `save_win_pos` and `arrange_windows`.
- Check if the PDF viewer is installed only when needed.
- Close #308


jalvesaq:
- Tested on a virtual machine.
- Couldn't try Wezterm as an external terminal because it fails with a message about missing OpenGL.
- Known bugs: SyncTeX forward and backward are not working (Rnoweb documents).

Can someone on a real Windows machine try it? @jasonpott, @kraxli, @dianchia, @DKBurns, @akthe-at, or @hongyuanjia?

PMassicotte:
No breaking changes on my side. Good job. Let's see if we can have a Windows user to try it?

jasonpott:
I'll attempt to load the commit now on my machine and report back. Are there a set of standard checks I should be running or just whether I run into any issues on start up?

jasonpott:
Commit 815591943c06ed03c0ece802ed2bf7a8279f993d installed without issue on win 10 in NVIM v0.11.1. 

QMD report runs, nvimcom loads completion lists for packages and functionality seems normal. 

Things seemed faster, often there is a lag before the R term can be started. Things were ready much quicker.

jalvesaq:
> No breaking changes on my side.

Thanks, @PMassicotte !

> Things seemed faster, often there is a lag before the R term can be started. Things were ready much quicker.

Thanks, @jasonpott ! Possibly, it's faster because of this change:

- Check if the PDF viewer is installed only when needed.

which is not directly related to the removal of `rgui.exe` support.

> Are there a set of standard checks I should be running

It's already a great contribution to run it as usual and report if it works, as you have already done.

On Linux, we rely on Tmux (which doesn't work on Windows) to run R in an external terminal emulator, a setup preferred by some of those working with two monitors. Maybe this is possible on Windows with Wezterm, but I can't try it because it doesn't run in my virtual machine. So, if you have spare time to investigate this possibility, it would be useful for some people.

jalvesaq:
Only now I remembered to increase nvimcom's version number, which is required to force its reinstallation.

jalvesaq:
I guess we should let Wezterm for a future pull request...

Playing with `kitty` and `kitten`, I see that there is no way to send messages to the terminal emulator without Tmux unless we explicitly support it.  For kitty, the command is, for example:

```
kitten @ send-text -m id:4 'ls()\n'
```

where `4` is the value of the environment variable `KITTY_WINDOW_ID` on the window where R is running, and `ls()` is the command to be sent. The only way of getting the environment variable is to make R evaluate the command `Sys.getenv("KITTY_WINDOW_ID")` and send the result back to `R.nvim`.

jasonpott:
I had to manually remove Nvimcom to get it to reinstall after your increment. 

Rendering a parameterized Quarto report worked when set_params = "no" but not when set_params = "yes". 
I didn't render the quarto report last night I just worked interactively.

The error message was:

```
processing file: ed-attendance-report.rmarkdown
1/13
2/13 [unnamed-chunk-1]
3/13
4/13 [unnamed-chunk-2]
Error:
! package or namespace load failed for 'nvimcom':
 .onLoad failed in loadNamespace() for 'nvimcom', details:
  call: assign("params", lapply(params, function(x) x$value), envir = .G
lobalEnv)
  error: cannot change value of locked binding for 'params'
Backtrace:
     ▆
  1. └─global .main()
  2.   └─execute(...)
  3.     └─rmarkdown::render(...)
  4.       └─knitr::knit(knit_input, knit_output, envir = envir, quiet =
 quiet)
  5.         └─knitr:::process_file(text, output)
  6.           ├─xfun:::handle_error(...)
  7.           ├─base::withCallingHandlers(...)
  8.           └─knitr:::process_group(group)
  9.             └─knitr:::call_block(x)
 10.               └─cache$library(params$cache.path, save = FALSE)     
 11.                 ├─base::suppressPackageStartupMessages(library(p, c
haracter.only = TRUE))
 12.                 │ └─base::withCallingHandlers(expr, packageStartupM
essage = function(c) tryInvokeRestart("muffleMessage"))
 13.                 └─base::library(p, character.only = TRUE)
 14.                   └─base::tryCatch(...)
 15.                     └─base (local) tryCatchList(expr, classes, pare
ntenv, handlers)
 16.                       └─base (local) tryCatchOne(expr, names, paren
tenv, handlers[[1L]])
 17.                         └─value[[3L]](cond)

Quitting from ed-attendance-report.rmarkdown:99-175 [unnamed-chunk-2]   
Execution halted
Error in `quarto::quarto_render()`:
✖ Error running quarto cli.
Caused by error:
! System command 'quarto.exe' failed
Run `rlang::last_trace()` to see where the error occurred.
```

>  Maybe this is possible on Windows with Wezterm, but I can't try it because it doesn't run in my virtual machine. So, if you have spare time to investigate this possibility, it would be useful for some people.

I use wezterm but I run R in an internal session, I will look at how to set up an external console for rnvim and see if I can get it working after work.

PMassicotte:
I agree with you @jalvesaq about wezterm in another PR.

PMassicotte:
> I guess we should let Wezterm for a future pull request...
> 
> Playing with `kitty` and `kitten`, I see that there is no way to send messages to the terminal emulator without Tmux unless we explicitly support it. For kitty, the command is, for example:
> 
> ```
> kitten @ send-text -m id:4 'ls()\n'
> ```
> 
> where `4` is the value of the environment variable `KITTY_WINDOW_ID` on the window where R is running, and `ls()` is the command to be sent. The only way of getting the environment variable is to make R evaluate the command `Sys.getenv("KITTY_WINDOW_ID")` and send the result back to `R.nvim`.

Do you see an advantage of using kitty directly over using tmux?

jalvesaq:
> Do you see an advantage of using kitty directly over using tmux?

It would be only one dependency less for Linux and macOS. But using Wezterm would be the only way to run R in an external terminal on Windows.

jalvesaq:
> Rendering a parameterized Quarto report worked when set_params = "no" but not when set_params = "yes".

I will install Quarto on the virtual machine...

PMassicotte (approved):

PMassicotte:
I think this can be merged?

jalvesaq:
> Rendering a parameterized Quarto report worked when set_params = "no" but not when set_params = "yes".

I could replicate the bug in the virtual machine, but it's fixed now.

jalvesaq:
> That looks good. If I understand correctly, with kitty we could "bypass" tmux usage?

Yes, we can add the integration with kitty, although it will not make a big difference because Linux and macOS users already have the same functionality with Tmux.

jalvesaq:
> I think this can be merged?

We can wait some time until @jasonpott can confirm that the `params` bug is fixed...

jasonpott:
OK I've updated to 73c23fc and the quarto reports now install on set_params "no" and "yes"

Thanks!

I've had  a look at the wezterm external term but will comment on the relevant [issue](https://github.com/R-nvim/R.nvim/issues/241)

jalvesaq:
Thanks! So, I think we can merge the pull request now.

PR #407, opened 2025-07-04
refactor: replace deprecated ts_utils with native vim.treesitter.get_node()

PMassicotte:
This pull request simplifies the code by replacing `ts_utils.get_node_at_cursor` with `vim.treesitter.get_node` in two files. This removes the dependency on `nvim-treesitter.ts_utils`.

The [`main` branch of nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter/tree/main) is now the actively maintained default branch, replacing `master`.  
The `ts_utils` module is no longer available in `main`, so this change ensures compatibility with the current API and avoids relying on deprecated functionality.

Could you try this is an R file with this content:

```r
read.csv("dir/path.csv")
```
with the cursor on the path and press `<localleader>sp`.

### DIFF
--- a/lua/r/format/brackets.lua
+++ b/lua/r/format/brackets.lua
@@ -71,16 +71,39 @@ M.formatsubsetting = function(bufnr)
 
     local node = vim.treesitter.get_node()
 
-    if node:type() == "extract_operator" then
+    local node_type = node:type()
+
+    if node_type == "extract_operator" then
         replace_extract_operator(node)
-    elseif node:type() == "identifier" then
+        return
+    end
+
+    -- Handle cursor on RHS identifier of extract operator (e.g., "var" in df$var)
+    if node_type == "identifier" then
         local parent = node:parent()
         if parent and parent:type() == "extract_operator" then
             replace_extract_operator(parent)
+            return
         end
-    elseif node:type() == "arguments" then
+    end
+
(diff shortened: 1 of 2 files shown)


jalvesaq:
`<LocalLeader>sp` works on both branches: "main" and "remove-usage-ts.utils".

`<LocalLeader>cn` also works (convert numbers to explicit integers).

However, `<LocalLeader>cs` isn't converting `dd$aa` into `dd[["aa"]]`.

PMassicotte:
> However, `<LocalLeader>cs` isn't converting `dd$aa` into `dd[["aa"]]`.

Do you have the cursor on `$`? For me it works.

PMassicotte:
I have modified the capture code a bit (76118d479e6331ef7c4786ce93d283c4db03da09)

You now have the cursor on the `$` and the `rhs` node. For example: `a$bbbb` and have the cursor on any `b`'s. Same for the `[]` operator, you can now have the cursor anywhere inside the `[]`.

jalvesaq:
It's working now with the cursor anywhere (over, before, or after the `$`).

jalvesaq:
The converstion of `d[1]` into `d[[1]]` also works. I think we can merge it.

PMassicotte:
Thank you!

PR #408, opened 2025-07-07
Integration with Wezterm and Kitty

### LINKED ISSUE #241: Support WezTerm mux splits

jalvesaq:
- Support for Wezterm and Kitty split panes (close #241).
- Don't check Tmux version (we require Tmux >= 3.0, released in 2019).
- Remove option `is_darwin`.


jalvesaq:
I tested the changes in combination with:

- `silent_term`: `true` and `false`.
- `clear_line`: `true` and `false`.
- `external_term`: "default", "wezterm", "wezterm_split", "kitty_split", "kitty", "foot", and "tmux split-window -vf".

It would be better to have confirmation that it works with all these options in macOS and a real Windows machine, but I think that it can be merged, and we can fix new bugs later.

PMassicotte:
Testing it now.

PMassicotte:
All good on my side. I can only test on Linux with Kitty. I tried both `kitty` and `kitty_split` and it works just fine. I think we can merge it. Thank you.

jalvesaq:
Thanks!

PMassicotte:
I like how you refactored debug, learned how you made it with `find_func` etc. :+1:

PR #410, opened 2025-07-24
Fix/chain edge case

PMassicotte:
This pull request refines the `M.chain` function in `lua/r/send.lua` to improve its logic for handling tree-sitter queries and streamline the code. The most notable changes involve simplifying the query structure, removing redundant checks, and ensuring the function behaves correctly based on cursor position when sending code.

Consider this example:

```r
list(
  mtcars |>
    head(1) |>
    slice.index(1)
)
```

Place the cursor on `head(1)` and try to send the chain with `:lua require('r.send').chain()`. I will incorrectly send the whole chain.

```r
mtcars |>
  head(1) |>
  slice.index(1)
```

 whereas it should be only sending:

```r
mtcars |>
  head(1)
```


jalvesaq:
Great! I confirm that the issue is fixed!

PR #411, opened 2025-08-03
Fix minor nvimcom bugs

jalvesaq:
- Add new known bug to doc/R.nvim.txt.
- Make library description available immediately after installation.
- Require R 4.1.0 due to `\(x)` in misc.R
- More robust nvimcom unload

### DIFF
--- a/nvimcom/src/apps/data_structures.c
+++ b/nvimcom/src/apps/data_structures.c
@@ -269,7 +269,7 @@ static PkgData *get_pkg(const char *nm) {
     return NULL;
 }
 
-static char *get_pkg_descr(const char *pkgnm) {
+static char *get_pkg_descr(const char *pkgnm, int again) {
     Log("get_pkg_descr(%s)", pkgnm);
     InstLibs *il = instlibs;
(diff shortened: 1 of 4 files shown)


jalvesaq:
## Add new known bug to doc/R.nvim.txt

The wrong tree segment is displayed in the Object Browser if the last element of a library is a kind of list. Example of affected package: `vdemdata`.

## Make library description available immediately after installation

The library title is not displayed in the Object Browser when the library is loaded immediately after being installed. Example:

install.packages("ineq")
library(ineq)

## Require R 4.1.0 due to `\(x)` in misc.R

The shortcut `\(x)` requires R 4.1.0.

## More robust nvimcom unload

R freezes with the following commands:

```r
detach("package:nvimcom", unload = TRUE)
library(nvimcom)
```

nvimcom/src/apps/data_structures.c
PMassicotte, line 1: Very nice!

PMassicotte (approved):
Good job. Works fine on my side. Thank you very much.

PR #412, opened 2025-08-05
fix roxygenize function

PMassicotte:
- **refactor(roxygen): simplify function parameter parsing using vim.treesitter API**
- **refactor: remove unused format.lua module for R number formatting (already defined used in lua/r/format/numbers.lua)**
- **refactor(roxygen): simplify imports and improve code formatting**


PMassicotte:
The roxygen changes are due to breaking changes in treesitter. The function `Roxygenize` now works again.

jalvesaq:
Thanks! In a quick test, the `:Roxygenize` command gives the same result on both the "main" and the "fix-roxygenize-function" branches.  So, the pull request simplifies the code without introducing new bugs. Great!

Note: all `R.nvim` commands start with two upper case letters, "R" and the second letter. "Roxygenize" is the only command with a lower-case second letter, "o". Some commands join two names and would be strange with a lowercase second letter, for example, "RDebugInfo", "RConfigShow", "RMapsDesc".

PMassicotte:
> Note: all `R.nvim` commands start with two upper case letters, "R" and the second letter. "Roxygenize" is the only command with a lower-case second letter, "o". Some commands join two names and would be strange with a lowercase second letter, for example, "RDebugInfo", "RConfigShow", "RMapsDesc".

Thanks for the review. Tomorrow I will make the changes.

Cheers

PMassicotte:
`Roxygenize` should be `ROxygenize` right?

jalvesaq:
> `Roxygenize` should be `ROxygenize` right?

It would be more consistent with the other names.

jalvesaq:
Thanks!

PR #419, opened 2025-08-18
Correction of README.md typo

jasonpott:
- In a Quarto or RMarkdown document, when the cursor is over a chunk header and you tries to send the current line and go to the next line, the whole  chunk is sent to R and the cursor jumps to the next chunk of either R or  Python code.

To

- In a Quarto or RMarkdown document, when the cursor is over a chunk header and you **try** to send the current line and go to the next line, the whole  chunk is sent to R and the cursor jumps to the next chunk of either R or  Python code.


PMassicotte:
Thank you!

PR #427, opened 2025-09-11
Check if label attribute is a character vector

### LINKED ISSUE #426: Issue with new version of ggplot2 (now using S7)
Steps to reproduce:

1. Install ggplot2 4.0.0
2. Create a plot and save it to a variable.
3. In the current buffer, type the name of that variable.

This will produce this error:

```
[ins] r$> Error in format_text(objlbl, " ", "\024") : 
  STRING_ELT() can only be applied to a 'character vector', not a 'list'
```

Demo:

Image

jalvesaq:
Close #426

### DIFF
--- a/nvimcom/DESCRIPTION
+++ b/nvimcom/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: nvimcom
-Version: 0.9.73
-Date: 2025-08-08
+Version: 0.9.74
+Date: 2025-09-11
 Title: Intermediate the Communication Between R and Neovim
 Author: Jakson Aquino
 Maintainer: Jakson Alves de Aquino <jalvesaq@gmail.com>
--- a/nvimcom/R/bol.R
+++ b/nvimcom/R/bol.R
@@ -379,7 +379,7 @@ nvim.cmpl.line <- function(x, envir, printenv, curlevel, maxlevel = 0) {
             } else {
                 info <- nvim.getInfo(printenv, x)
                 if (info == "\006\006") {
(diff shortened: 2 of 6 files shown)


jalvesaq:
Please, try this:

```r
attr(p, "label")
attr(p, "label", exact = TRUE)
```

I'll improve the pull request...

jalvesaq:
After this code:

```R
library(ggplot2)
p <- ggplot(mtcars, aes(wt, mpg)) +
  geom_point()
```

Typing `p@` triggers the completion menu in RStudio. However, I can't implement the same functionality in `R.nvim` because I need the C function `Rf_isS7`, which doesn't exist yet.

Is there anything else that we can do to improve support for `S7` objects? If not, I believe the pull request can be merged.

PMassicotte:
I think this is a very good start. Do you think we could vendor `Rf_isS7`?

jalvesaq:
This seems to work:

```c
int Rf_isS7(SEXP x) {
    SEXP cls;
    int isS7 = 0;
    PROTECT(cls = getAttrib(x, R_ClassSymbol));
    if (!isNull(cls)) {
        for (int i = 0; i < Rf_length(cls); i++) {
            if (strcmp(CHAR(STRING_ELT(cls, i)), "S7_object") == 0) {
                isS7 = 1;
                break;
            }
        }
    }
    UNPROTECT(1);
    return isS7;
}
```

jalvesaq:
Much simpler:

```c
int isS7 = Rf_inherits(x, "S7_object")
```

PR #433, opened 2025-09-20
fix(health.lua): ensure parser_installed checks for has_parser function availability

PMassicotte:
Improved robustness of the `parser_installed` function by adding a check for the existence of `treesitter_parsers.has_parser` before invoking it.

### DIFF
--- a/lua/r/health.lua
+++ b/lua/r/health.lua
@@ -8,8 +8,11 @@ local ts_available, treesitter_parsers = pcall(require, "nvim-treesitter.parsers
 ---@param parser_name string
 ---@return boolean
 local function parser_installed(parser_name)
-    return (ts_available and treesitter_parsers.has_parser(parser_name))
-        or pcall(vim.treesitter.query.get, parser_name, "highlights")
+    return (
+        ts_available
+        and treesitter_parsers.has_parser
+        and treesitter_parsers.has_parser(parser_name)
+    ) or pcall(vim.treesitter.query.get, parser_name, "highlights")
 end
 
 --- Create a buffer and check the languages at different cursor positions


PMassicotte:
This issue arise depending on whether the user uses the main or the deprecated master branch of treesitter.

jalvesaq:
Thanks!

How about the other minor issue (`:checkhealth r` reports two language detection errors if a Markdown/Quarto/Rmd file was not opened yet)? If we can't fix it, we should include it in the _Known bugs_ section.

PMassicotte:
I will have a look.

PMassicotte:
Can you try to run `:checkhealth r` twice in a row? For me, the second time, all is good.

image

jalvesaq:
I'm in the process of migrating to the "main" branch of `nvim-treesitter` and starting to use `vim.lsp.enable()` and `vim.lsp.config()`. Something might be wrong in my config, but, currently, I had to make this change to get `:checkhealth r` working:

```diff
diff --git a/lua/r/health.lua b/lua/r/health.lua
index 9b080d4..398870c 100644
--- a/lua/r/health.lua
+++ b/lua/r/health.lua
@@ -35,6 +35,7 @@ local check_buffer = function(ft, lines, langs)
 
     vim.api.nvim_buf_set_lines(b, 0, -1, false, lines)
     vim.api.nvim_set_option_value("filetype", ft, { buf = b })
+    vim.treesitter.start(b, "markdown")
     vim.cmd("redraw")
 
     for k, v in pairs(langs) do
```

PMassicotte:
Do you mean that adding this fix the errors when running `:checkhealth r`? Because for me, I do not need this, but I have check errors.

PMassicotte:
https://github.com/R-nvim/R.nvim/pull/434 if fixing the 2 errors we add.

jalvesaq:
Adding this, the first run of `checkhealth r` gives the two language errors, and the second run works without reporting errors, as in your case.

PR #439, opened 2025-09-24
Update references to Nvim-R (now Vim-R)

jalvesaq:


jalvesaq:
In addition to updating the references to Nvim-R (which now is Vim-R), I also removed the check of conflicting installations from `config.lua` because now Vim-R disables itself if R.nvim is installed (which will make the startup a few nanoseconds faster). I kept the check in `health.lua` because it will only run when the user manually executes `:checkhealth r`.

PR #440, opened 2025-09-25
feat(quarto_rnvim): add YAML LSP configuration and Quarto schema

PMassicotte:
This should do it for a first experimental version.

The `/opt/quarto/share/editor/tools/yaml/yaml-intelligence-resources.json` and `/opt/quarto/share/editor/tools/yaml/yaml-intelligence.js` are quite substantial. The completion provided in RStudio is more complex and relies on Treesitter.

I went down the rabbit hole trying to get proper yaml schema from `/opt/quarto/share/editor/tools/yaml/yaml-intelligence-resources.json` using LLM. I think this will now provide basic and most common completion items in `_quarto.yml` and `*.qmd` front matters.

We can modify and adjust the `quarto-project-schema.json` and `quarto-schema.json` when we discover missing keys.

### DIFF
--- a/ftplugin/yaml_rnvim.lua
+++ b/ftplugin/yaml_rnvim.lua
@@ -1,4 +1,3 @@
--- ftplugin/yaml.lua - Handle _quarto.yml files
 local filename = vim.fn.expand("%:t")
 
 if filename == "_quarto.yml" or filename == "_quarto.yaml" then


jalvesaq:
Testing here... Everything seems to be good. The only thing is that I suppose we don't need a project schema because the user might want to define everything in the `_quarto.yml` file if there is a single `.qmd` file in the directory. It's also valid to define almost everything, leaving for the YAML metada of each `.qmd` file only what is specific to that document.

jalvesaq:
Looking at the output of `:checkhealth r`, I see that the result of the `yaml-language-server` executable check is in the category "Checking language detection in an Rnoweb document" which isn't exactly correct. Perhaps we should add the result to the "Checking applications and plugins" category.

PMassicotte:
> Testing here... Everything seems to be good. The only thing is that I suppose we don't need a project schema because the user might want to define everything in the `_quarto.yml` file if there is a single `.qmd` file in the directory. It's also valid to define almost everything, leaving for the YAML metada of each `.qmd` file only what is specific to that document.

I am not sure to understand. You would prefer to drop completion in `_quarto.yml` with elements in `quarto-project-schema.json`?

PMassicotte:
> Looking at the output of `:checkhealth r`, I see that the result of the `yaml-language-server` executable check is in the category "Checking language detection in an Rnoweb document" which isn't exactly correct. Perhaps we should add the result to the "Checking applications and plugins" category.

Fixed in 9d652839b7f0

jalvesaq:
Calling `dofile()` might not be the best option because it doesn't cache the file contents. It might be better to call `require()`.

We can't `require` `quarto.lua` from `ftplugin/yaml_rnvim.lua` because this will load all of R.nvim functionality, including the ability to start R from a Yaml file. So, I think it's better to refactor `lua/r/quarto.lua` and move the YAML code to a new `lua/r/yaml.lua`.

jalvesaq:
> I am not sure to understand. You would prefer to drop completion in _quarto.yml with elements in quarto-project-schema.json?

I prefer to use only the biggest json file (`quarto-schema.json`) for both Quarto documents and `_quarto.yml`. Is there is anything in `quarto-project-schema.json` that is not also in `quarto-schema.json`? If, yes, then, the two could be merged to include everything in `quarto-schema.json`.

PMassicotte:
> If, yes, then, the two could be merged to include everything in `quarto-schema.json`.

Can you use the `project` key (for example) in a .qmd file? I thought it was only usable in `_quarto.yml`. If it can be used in both filetypes, your suggestion makes sense.

I am trying to use `project` config in a single .qmd file

```yml
---
project:
  type: website
  output-dir: docs
  toc: false
  toc-depth: 2
  toc-location: left
  theme: darkly
  navbar:
    right:
      - text: "About"
        href: about.qmd
      - text: "Posts"
        href: posts/posts.qmd
      - text: "GitHub"
---
```

But this is not producing a website. Not sure if completion for `project` should be available in a single qmd file. This is why I used to `json`.

Maybe we can just remove the `project` key from the json file and have a single json as you said.

PMassicotte:
image
Apparently yes...

jalvesaq:
> But this is not producing a website. Not sure if completion for project should be available in a single qmd file. This is why I used to json.

I didn't know about that because I have never used Quarto projects. I had just tested to insert "title" and "author" in the `_quarto.yml` and it worked.

PMassicotte:
I am brain-dead for today I will continue on that tomorrow and think what is the best solution ;)

jalvesaq:
Thanks!

PMassicotte:
After carefully thinking and searching, I could not find a good way to only have 1 json schema file. There are indeed some completion that are only project specific (i.e that can only be used in `_quarto.yml`, https://quarto.org/docs/projects/quarto-projects.html#project-metadata). 

If we want to support both project and file specific we have to use two different schemas. This is how Redhat does it with `yamlls` (https://github.com/redhat-developer/yaml-language-server).

```yaml
"yaml.schemas": {
    "http://json.schemastore.org/composer": ["/*"],
    "file:///home/johnd/some-schema.json": ["some.yaml"],
    "../relative/path/schema.json": ["/config*.yaml"],
    "/Users/johnd/some-schema.json": ["some.yaml"],
}
```
I never tried that, but if we go with 2 files, there is a way to combine shared references. 

https://tour.json-schema.org/content/06-Combining-Subschemas/01-Reusing-and-Referencing-with-defs-and-ref

jalvesaq:
So, we would have one file for Quarto and two for YAML...

jalvesaq:
I'll be able to review it again only at night...

PMassicotte:
> So, we would have one file for Quarto and two for YAML...

Just 2, one for things going in `_quarto.yml` and another for `*.qmd` files.

jalvesaq:
Currently, the two json files have repeated entries. This is a maintenance problem because bugs would have to be fixed in two place, new entries added to two files, deprecated entries deleted from two files...

So, would this work:

in `lua/r/yaml.lua`:

```lua
            schemas = {
                ["file://" .. document_schema] = "*.qmd",
                ["file://" .. document_schema] = { "_quarto.yml", "_quarto.yaml" },
                ["file://" .. project_schema] = { "_quarto.yml", "_quarto.yaml" },
            },
```

The document schema would include almost everything and the project schema would include only what is project specific.

PMassicotte:
Yeah!

jalvesaq:
Thinking about the most appropriate directory to put the JSON files. They are not Lua files, so `lua/r/` might not be the correct place for them. They are not scripts either. So, perhaps we should rename the `scripts` directory as `resources` and put the JSON files there. What do you think?

PMassicotte:
Yes , this is a good idea. 

I am trying your idea, slightly differently since we can't have duplicated keys in lua table.

```lua
chemas = {
                ["file://" .. document_schema] = {
                    "*.qmd",
                    "_quarto.yml",
                    "_quarto.yaml",
                },
                ["file://" .. project_schema] = { "_quarto.yml", "_quarto.yaml" },
            },
```

PMassicotte:
> Thinking about the most appropriate directory to put the JSON files. They are not Lua files, so `lua/r/` might not be the correct place for them. They are not scripts either. So, perhaps we should rename the `scripts` directory as `resources` and put the JSON files there. What do you think?

2e81f2d9725f

jalvesaq:
Thanks!

Just a typo: I agree that "resources" should be spelled as "ressources", but English has very irregular spelling and pronunciation.

PMassicotte:
I will add new tests that will make sure we do not have duplicated keys in both json and then should be good to go

jalvesaq:
I have one last question: If a user already has a configuration for the YAML server, will our configuration overwrite the existing one? Should we try to merge the two configurations? Anyway, we may leave this for after merging the pull request, if anyone reports any bug.

PMassicotte:
> I will add new tests that will make sure we do not have duplicated keys in both json and then should be good to go

That is a good question... I will do some tests.

PMassicotte:
> I have one last question: If a user already has a configuration for the YAML server, will our configuration overwrite the existing one? Should we try to merge the two configurations? Anyway, we may leave this for after merging the pull request, if anyone reports any bug.

I did that to keep only the buffer-local lsp starting

https://github.com/R-nvim/R.nvim/blob/6dd86919bcab74cb3952ccdc0807c627f12afcb6/lua/r/yaml.lua#L34-L40

Can you confirm it still work on your side?

PMassicotte:
I have added lint check on both schema, this will fail if we have invalid json or duplicated keys.
 
8268465340f1

jalvesaq:
Good job! This completes the planed features for 1.0.

PR #441, opened 2025-09-28
Fix yaml_hl

jalvesaq:
When trying to highlight YAML comments and Jupyter delimiters in RMarkdown and Quarto documents, we are linking them to non-default tree-sitter highlight groups: `@include` and `@delimiter`. Unless the user has defined these highlight groups, the highlighting would fail. This pull request changes the highlight groups to `@attribute` and `@string.special`. I tryied it with the color schemes "habamax" and "tokyonight". Example of code to test the highlighting:

````rmd

Normal text here.

```{r}
#| output: false
#| echo: false

# Normal comment

aa <- rnorm(100)
```

Normal text here.

```{python}
#| label: setxvalue

#%% This is a Jupyter delimiter

x = 10
print(x)
```

````

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -1778,13 +1778,13 @@ config, as in the examples:
 R.nvim changes the background color of code blocks and adds a virtual text
 with the language of the code block to the right side of the header of the
 block. It also highlights YAML comments (prefixed with `#|`) with the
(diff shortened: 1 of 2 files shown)


PMassicotte (approved):
Nice improvement, switching to standard groups to ensures better highlighting across color schemes!

PMassicotte:
Looks better!

image

PR #443, opened 2025-10-10
feat(nvimcom): add names completion for s7 objects

jamespeapen:
This provides support for S7 property completion with the `@` trigger. I saw that S7 support was added to cmp-r and tried using S7 but `nvimcom::nvim.names` wasn't completing the slot/properties with the `@` trigger because `names()` is not a method for S7 objects. Using the object attributes names nvimcom can access S7 properties without having to load the S7 package.

If you're waiting for S7 to be included in base R feel free to close.

### DIFF
--- a/nvimcom/R/misc.R
+++ b/nvimcom/R/misc.R
@@ -429,7 +429,7 @@ nvim.plot <- function(x) {
 nvim.names <- function(x) {
     if (isS4(x)) {
         slotNames(x)
-    } else if ("S7_object" %in% class(x)) {
+    } else if (inherits(x, 'S7_object')) {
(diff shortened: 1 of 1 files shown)


PMassicotte:
LGTM. I see no reason to not merge it. @jalvesaq ?

jalvesaq:
I didn't try it, but it's a simple change, and looks good. Thank you!
Note: without increasing nvimcom version, the nvimcom package will not be automatically upgraded.

jalvesaq:
I'll make a pull requesting fixing some linting warnings and will take the opportunity to increase the nvimcom version...

PR #454, opened 2025-11-18
Feature/formula column completion in ggplot2

PMassicotte:
What I have used to test

```r
ggplot(mtcars, aes(x = carb)) +
  geom_bar() +
  facet_wrap(cyl ~ mpg)


mtcars |>
  ggplot(aes(x = carb)) +
  geom_bar() +
  facet_grid(~mpg)
```

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -1517,7 +1517,7 @@ Meaning of optons:
 
     Example usage:
 >r
-    ggplot(mtcars, aes(x = mpg)) +
+    ggplot(mtcars, aes(x = mpg, y = hp)) +
         geom_point() +
         facet_wrap(~cyl)  # Typing ~ will complete column names from mtcars
 <


PMassicotte:
I noticed that I am not getting completion in `ggplot(mtcars, aes(x = carb))` when I type `x = `. Did I messed up something?

jalvesaq:
That's great! Using tree-sitter is better than my approach of using regular expressions. I have to leave for work now. I'll be able to look at this tonight or, perhaps, only on Thursday.

jalvesaq:
> I noticed that I am not getting completion in `ggplot(mtcars, aes(x = carb))` when I type `x = `. Did I messed up something?

The completion works on my side. I always have to press `<C-Space>` to manually trigger completion of arguments with Blink, but the `mtcars` columns are included in the menu after the function arguments.

jalvesaq:
> What I have used to test

It works! The only problem is the example in the documentation. The `y` argument is missing from the `aes()` function and `geom_point()` requires it.

PMassicotte:
> It works! The only problem is the example in the documentation. The `y` argument is missing from the `aes()` function and `geom_point()`

Oups, fixed!

jalvesaq:
Thank you!

PR #458, opened 2025-11-22
Ensure the :checkhealth window gets the focus when the temporary buffer is closed

### LINKED ISSUE #457: When R.nvim is enabled :checkhealth does not show lsp configuration
This issue is probably related to #455. After I have updated to the latest commit, the following issue has occured.

**Current Behavior:**
When running `:checkhealth` in an environment where no R file is opened, the cursor starts in a weird position below the health window directly above the line where the vim mode is displayed.

Image

You can just click on the health window to navigate it. But I also noticed that the vim.lsp output isn't displayed anymore in the general health overview. I tried running `:checkhealth vim.lsp` once and it crashed. Running it again has produced the expected output so the crash might be unrelated.

**Expected Behavior / Behavior when R.nvim is disabled:**
- The cursor is placed inside the actual health overview
- The vim.lsp health output is displayed in the general health overview

johannesbzd:
This pull request fixes #457 by ensuring that the original health window gets the focus as soon as the temporary window which is used to check the functionality of the tree-sitter parsers is closed. Please note the issue does not seem to be reproducible on other machines but I don't see how this change would have an impact on other functioning setups.

### DIFF
--- a/lua/r/health.lua
+++ b/lua/r/health.lua
@@ -15,116 +15,6 @@ local function parser_installed(parser_name)
     ) or pcall(vim.treesitter.query.get, parser_name, "highlights")
 end
 
---- Create a buffer and check the languages at different cursor positions
----@return table
-local check_buffer = function(ft, lines, langs)
-    local health_win = vim.api.nvim_get_current_win()
-
-    -- Create a temporary buffer in a temporary window
-    local b = vim.api.nvim_create_buf(false, false)
-    vim.api.nvim_set_option_value("bufhidden", "wipe", { scope = "local" })
-    vim.api.nvim_set_option_value("number", false, { scope = "local" })
-    vim.api.nvim_set_option_value("swapfile", false, { scope = "local" })
-    vim.api.nvim_set_option_value("buftype", "nofile", { scope = "local" })
-    local w = vim.api.nvim_open_win(b, true, {
-        relative = "win",
-        row = 1,
-        col = 1,
-        width = 20,
-        height = 13,
-        hide = true,
-    })
-
-    vim.api.nvim_buf_set_lines(b, 0, -1, false, lines)
-    vim.api.nvim_set_option_value("filetype", ft, { buf = b })
-    vim.cmd("redraw")
-
-    -- Wait for treesitter to parse the buffer, likely a racing issue here
-    vim.wait(100)
-
-    for k, v in pairs(langs) do
-        vim.api.nvim_win_set_cursor(w, { v[2], v[3] })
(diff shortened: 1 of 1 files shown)


jalvesaq:
Thank you! I tried your branch, and now I get this:

```
Checking language detection in a Quarto document: ~
- ❌ ERROR Wrongly detected: `yaml` vs `none`
- ❌ ERROR Wrongly detected: `markdown` vs `none`
- ❌ ERROR Wrongly detected: `markdown_inline` vs `none`
- ❌ ERROR Wrongly detected: `r` vs `none`

Checking language detection in an Rnoweb document: ~
- ❌ ERROR Wrongly detected: `latex` vs `none`
- ❌ ERROR Wrongly detected: `chunk_header` vs `none`
- ❌ ERROR Wrongly detected: `r` vs `none`
- ❌ ERROR Wrongly detected: `chunk_end` vs `none`
```

It seems that we need the `enter` window parameter to be `true`.

johannesbzd:
I get the same error messages, didn't notice them on my first run because I only checked whether the information of the other plugins are displayed. But when the temporary window gains focus on creation the issue persists and I don't get any output from other plugins or R.nvim itself in the checkhealth overview

jalvesaq:
So, the bug persists? We may move these checks to `:RDebugInfo`, avoiding the need to create a buffer during `:checkhealth`.

johannesbzd:
Yes at the current state of this branch the bug persists although the focus should be moved to the checkhealth window after the tree-sitter parsers have been checked.

To be exact, currently ...:
- The cursor is still in the weird position at the start of checkhealth
- The outputs of many plugins are missing

jalvesaq:
So, please, delete the offending code. I'll bring it back to another place later.

johannesbzd:
Sorry, english isn't my first language and I'm not sure what you mean with "offending". The code which caused the error messages is already deleted. The only change which remains is the change of focus after the tree-sitter parsers have been checked which doesn't seem to have any effect or solve the issue #457.
Can you please tell me which code exactly should be removed?

jalvesaq:
English isn't my first language either. So, "offending" might not be the appropriate word.

Just delete the line

```
    check_lang()
```

near the end of `health.lua`. Then, delete the local functions that will no longer be used: `check_buffer()`, `print_lang_check()`, and `check_lang()`. This way, your contribution will be recorded in the git log.

johannesbzd:
The functions and the invocation have been removed. Thanks for the guidance!

jalvesaq:
Thank you!

PR #459, opened 2025-11-22
Always try to show information on hover

jalvesaq:


jalvesaq:
@PMassicotte, I'm making the pull request, but you might want to improve it with more intelligent tree-sitter handling...

PMassicotte:
Great, I think `get_captures_at_cursor()` was deprecated in favour of  `get_captures_at_pos()`. I will propose some minor changes.

PMassicotte:
Made some minor changes, what do you think?

jalvesaq:
It's better! The formula completion isn't working, but that's because we didn't rebase on main.

PMassicotte:
One thing we could improve:

```r
mtcars$am
```

It currently gives the same results whether we are on `mtcars` or `am`. We could grab exactly the right thing. Let me check that.

PMassicotte:
> Let me check that.

I think that would require substantial changes in 

https://github.com/R-nvim/R.nvim/blob/8f6c810a70895d3851182a414e85508d58fc1808/lua/r/cursor.lua#L228 

We can do that later when we switch for a full TS based solution. Maybe we can merge for now.

jalvesaq:
I agree.

PR #460, opened 2025-11-22
Check language detection in `:RDebugInfo`

jalvesaq:


jalvesaq:
Bring back the language detection code, now as part of `:RDebugInfo`.

PMassicotte:
Perfect!

PR #464, opened 2025-12-01
Fix 3 bugs:

jalvesaq:
- `<Enter>` or double click on `data.frame` in the Object Browser doesn't expand the `data.frame`.
 - `<Enter>` or double click on a library's function in the Object Browser fails to open the function documentation.
 - Hover over `graphics::plot` (or any `package::function`) doesn't work.


PMassicotte:
Right on, fixed the issues on my side. Thank you.

PR #470, opened 2025-12-11
fix(send.lua): update string formatting for Python code execution to include delimiters for clarity and consistency (closes #469)

### LINKED ISSUE #469: Use robust raw string syntax `r"---(...)---"` for sending Python chunks
When sending Python chunks to the R console (e.g., via `reticulate::py_run_string`), the current implementation  uses standard raw strings `r"(...)"`. This is fragile because if the Python code contains the character sequence `)"`—which is common in print statements or string formatting—R interprets it as the end of the raw string, causing a syntax error.

**Example Failure:**
Given this Python chunk in Quarto:

````python
```{python}
print("(hello)")
print("File size (1MB)")
```
````

Sending this to the console (via `<Enter>` or `\cc`) results in:

```r
> reticulate::py_run_string(r"(print("(hello)"))")
Error: unexpected ')' in "reticulate::py_run_string(r"(print("(hello)"))"
```

Using `---`(or more dashes) makes the wrapper significantly more robust against code containing `)"`.

**Demonstration:**

```r
# Current (Fails)
> reticulate::py_run_string(r"(print("(hello)"))")
Error: unexpected ')' in "reticulate::py_run_string(r"(print("(hello)"))"

# Proposed (Works)
> reticulate::py_run_string(r"---(print("(hello)"))---")
(hello)
```

PMassicotte:
See #469.

Thanks to @jiangyun-fun for the suggestion.

### DIFF
--- a/lua/r/send.lua
+++ b/lua/r/send.lua
@@ -212,7 +212,7 @@ M.source_lines = function(lines, what)
     if #lines < config.max_paste_lines then
         rcmd = table.concat(lines, "\n")
         if what and what == "PythonCode" then
-            rcmd = 'reticulate::py_run_string(r"(' .. rcmd .. ')")'
+            rcmd = 'reticulate::py_run_string(r"---(' .. rcmd .. ')---")'
         end
     else
         vim.fn.writefile(lines, config.source_write)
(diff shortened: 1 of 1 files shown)


jalvesaq:
Thank you!

PR #475, opened 2025-12-15
chore(ci): update CI configuration to install additional Lua dependencies and ensure the treesitter cli is installed

PMassicotte:
Fix issues due to treesitter that has officially switched from the master to the main branch. Making sure that all parsers are automatically installed and can be found in the PATH.


jalvesaq:
Thank you!

PR #449, opened 2025-11-06
Turn rnvimserver into a language server

jalvesaq:

PR #478, opened 2025-12-20
Fix missing functions in help.R

### LINKED ISSUE #477: Error on help: `'is_dev_pkg' is not an exported object`
When running `<leader>rh` (help for function which doesn't exist) I get the error: 

`Error: 'is_dev_pkg' is not an exported object from 'namespace:pkgload'`

I think this line needs changing (`pkgload::is_dev_package` does exist: https://pkgload.r-lib.org/reference/is_dev_package.html)

https://github.com/R-nvim/R.nvim/blob/d8eeaabfe74acb034eacf3793c459609b9fe7ebe/nvimcom/R/help.R#L64

Thanks for the impressive development over the past few months on R.nvim!

wklimowicz:
Fixes #477: running help on functions correctly returns the help, without errorr-ing out on the if statements.

This was introduced by #449 when stuff was renamed from `package` to `pkg`.

### DIFF
--- a/nvimcom/DESCRIPTION
+++ b/nvimcom/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: nvimcom
-Version: 0.9.80
-Date: 2025-12-12
+Version: 0.9.81
+Date: 2025-12-20
 Title: Intermediate the Communication Between R and Neovim
 Author: Jakson Aquino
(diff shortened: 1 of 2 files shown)


jalvesaq:
Thank you!

PR #485, opened 2026-01-08
Avoid bug when slot value is NULL (close #481)

### LINKED ISSUE #481: `ggplot2` (>= 4.0.0) objects break Object Browser (S7/S4 slot error)
Happy New Year!

## Description

Creating certain `ggplot2`-based objects (specifically via `ggVennDiagram`) triggers a slot-related error. This error causes the **Object Browser** to stop updating. Subsequent variables defined in the session (e.g., `foo`) do not appear in the browser tree.

## Minimal Reproducible Example

```r
library(ggVennDiagram)

# 1. Create data
x <- list(A = 1:5, B = 2:7, C = 3:6, D = 4:9)

# 2. This triggers the error in the R.nvim environment
# The error likely occurs when the background callback 'cb' inspects 'p'
p <- ggVennDiagram(x)

# 3. This variable will NOT appear in the Object Browser
foo <- "foo"

```

## Observed Error

```log
Error in cb(expr, value, succeeded, visible) : 
  no slot of name "family" for this object of class "ggplot2::element_text"

```

## Supplement

* The issue occurs on `ggplot2 4.0.1`.
* The issue **disappears** when rolling back to `ggplot2 3.5.1`.

## Video Evidence

https://github.com/user-attachments/assets/cd05556f-1baa-4834-bfce-07187ce17ee4

jalvesaq:


PMassicotte:
Thank you, fixed the issue on my side also.

PR #486, opened 2026-01-19
Fix completion of package::function

jalvesaq:
@gueyenono reported a limitation of R.nvim at https://github.com/R-nvim/R.nvim/discussions/479

This pull request overcomes this limitation and makes completion, signature, and hover work properly with functions prefixed with their library names (`package::function`) even when the libraries were never loaded.

Previously, only loaded libraries had their data kept in rnvimserver's memory. Now, there are two lists: one with the data of all installed libraries and the other with pointers to the data of loaded libraries. Previously, only libraries loaded at least once had their cache files generated. Now, we need to generate the cache data for all libraries, including libraries that we never loaded directly but that other packages depend on. Cache files are generated in parallel to compensate for the increased number of libraries whose data has to be cached.


PMassicotte:
Testing it!

PMassicotte:
I see package completion going very fast! And everything is set up in `/home/xxx/.cache/R.nvim/objls_nvimcom_0.9.83`. However, I do not get completion.

```r
[ins] r$> packageVersion("nvimcom")
[1] ‘0.9.83’
```

Investigating.

jalvesaq:
> However, I do not get completion.

Don't you get normal completion (without the `package::` prefix)? As it was before the changes in this pull request, only the functions of loaded packages are expected to be completed.

PMassicotte:
> > However, I do not get completion.
> 
> Don't you get normal completion (without the `package::` prefix)? As it was before the changes in this pull request, only the functions of loaded packages are expected to be completed.

I do get normal completion (as before). 

Do you mean that I should load, for example, `fs` to be able to use `::`?

Like

```r
library(fs)

fs:: 

```

PMassicotte:
Peek 2026-01-19 11-37

jalvesaq:
The exact way the completion works depends on the completion plugin. `nvim-cmp` is more aggressive and constantly restarts the completion. With `blink.cmp`, you have to restart the completion manually after typing `::`.

PMassicotte:
> With `blink.cmp`, you have to restart the completion manually after typing `::`.

This is what I did, the completion is triggered with `ctrl-space` here. But it is only showing buffer completion.

PMassicotte:
I am getting signature completion now even if the library is not loaded.

image

jalvesaq:
> This is what I did, the completion is triggered with ctrl-space here. But it is only showing buffer completion.

i confirm that `fs::<c-space>` doesn't work. I have to type at least an additional letter: `fs::a<c-space>`.

PMassicotte:
Ah yes, same!

PMassicotte:
I found that I have to exit insert mode, re-enter it to make it pop:

Peek 2026-01-19 12-12

jalvesaq:
> I found that I have to exit insert mode, re-enter it to make it pop:

Yes, that's a different way of restarting the completion. It has the same effect as  `<c-space>`.

The list of trigger characters already includes `:`, so I don't know what else we could do to automatically restart the completion after `::`.

PMassicotte:
I can try to look at this tonight or tomorrow morning.

jalvesaq:
It works with this change:

```diff
diff --git a/nvimcom/src/apps/complete.c b/nvimcom/src/apps/complete.c
index 30ec97b..dce6a6b 100644
--- a/nvimcom/src/apps/complete.c
+++ b/nvimcom/src/apps/complete.c
@@ -129,7 +129,7 @@ static char *parse_objls(const char *s, const char *base, const char *pkg,
 
     while (*s != 0) {
         int z = fuzzy_find(s, base);
-        if (z) {
+        if (*base == '\0' || z) {
             i = 0;
             while (i < 7) {
                 f[i] = s;
```

which means: add the item to the completion menu if there is no "base" (the characters already typed) or the "base" can be found through the item (fuzzy finding).

PMassicotte:
ah super! works on my side too!

PMassicotte:
Should the completion also propose namespace as we type, like showing `tidyverse` when typing `tidy..`

jalvesaq:
> Should the completion also propose namespace as we type, like showing tidyverse when typing tidy..

This basically would mean to add them to all completion menus... Maybe it should be optional, since most people load the libraries instead of using double colons to call specific functions.

PMassicotte:
We can go like this for now (your PR) and maybe add package completion later?

PMassicotte:
Peek 2026-01-19 14-06

Awesome work!

jalvesaq:
image

I added the package completion (see above), but `blink.cmp` is not respecting the `sortText` field. As you can see from the log, the "zz" prefixed to "dplyr" had no effect:

```
SEND_LS_RESPONSE (301 bytes):
{"jsonrpc":"2.0","id":2,"result":{"isIncomplete":false,"items":[{"label":"dendrapply","sortText":"10dendrapply","cls":"F","kind":3,"env":"stats"},{"label":"duplicated.array","sortText":"16duplicated.array","cls":"F","kind":3,"env":"base"},{"label":"dplyr::","sortText":"zzdplyr","cls":"L","kind":9}]}}
```

jalvesaq:
image

The menu is correct with `nvim-cmp`.

PMassicotte:
I do not have this error on my side (or I misunderstood?)

Peek 2026-01-19 14-36

PMassicotte:
I think I misunderstood your point. What do you mean, it is not respecting the order? The only difference is that `blink`, `dplyr` is at the top of the list?

jalvesaq:
It might be my blink config. I'm not at home now, but I can look at this in about an hour...

jalvesaq:
> The only difference is that blink, dplyr is at the top of the list?

Yes, this was the problem, but it is fixed by this `blink` configuration:

```lua
    fuzzy = {
        sorts = {
            "exact", -- Prioritize exact matches first
            "sort_text", -- Then use the LSP's sortText
            "score", -- Fallback to fuzzy matching score
            "label", -- Finally, sort by the display label
        },
    },
```

I think it can be merged now.

jalvesaq:
I think I made a mistake, forcing pushing the wrong commit...

PR #490, opened 2026-01-22
fix: prevent infinite loop in get_word with emoji characters

### LINKED ISSUE #489: Neovim freezes when typing after emoji in .qmd/.R files
## Summary

Neovim completely freezes (CPU 100%, requires `kill -9`) when:
1. Opening a `.qmd` or `.R` file containing emoji (e.g., 📊)
2. Pressing `a` to enter append mode after the emoji
3. Typing a space
4. Typing any character → immediate freeze

## Reproduction Steps

```bash
# Test file content
echo "📊 " > test.qmd

# Open with R.nvim
nvim test.qmd

# Reproduce freeze:
# 1. Press 'a' (append mode)
# 2. Press space
# 3. Type any character → FREEZE
```

## Root Cause

The `get_word()` function in `lua/r/lsp/init.lua` has a bug:

**Problem**: LSP returns character positions (0-based), but Lua's `line:sub()` uses byte positions.

For emoji like 📊 (4 bytes in UTF-8, 1 character):
- When cursor is at position 4 (character), `vim.fn.byteidx("📊 ", 4)` returns `-1` (beyond string)
- Original code set `i = -1`, causing infinite loop
- The loop never reached its break condition

```lua
-- BEFORE (buggy)
local get_word = function(line, cnum, pttrn)
    local i = cnum  -- ❌ cnum is character position, not bytes!
    -- ... infinite loop with emoji
end
```

```lua
-- AFTER (fixed)
local get_word = function(line, cnum, pttrn)
    local byte_idx = vim.fn.byteidx(line, cnum)  -- ✅ Convert to bytes
    if byte_idx < 0 then
        byte_idx = #line  -- Handle cursor beyond string
    end
    local i = byte_idx
    -- ... works correctly
end
```

## Fix

1. Convert LSP character position to byte position using `vim.fn.byteidx()`
2. Handle `byteidx` returning `-1` when cursor is beyond string end
3. Add iteration safety limit (100) as additional protection

jiangyun-fun:
When typing after multi-byte characters (emoji like 📊), Neovim would freeze. The get_word function in lua/r/lsp/init.lua incorrectly used LSP character position as byte position when calling line:sub().

Changes:
- Convert character position to byte position using vim.fn.byteidx()
- Handle byteidx returning -1 when cursor is beyond string end
- Add iteration safety limit to prevent infinite loops

(close #489)

### DIFF
--- a/lua/r/lsp/init.lua
+++ b/lua/r/lsp/init.lua
@@ -337,38 +337,9 @@ end
 ---@param cnum integer Cursor position (LSP: UTF-16 code units, same as characters for BMP)
 ---@param pttrn? string Pattern to get word for completion
 local get_word = function(line, cnum, pttrn)
-    -- Convert character position to byte position for line:sub()
-    -- byteidx returns -1 if cnum is beyond the string length
     local byte_idx = vim.fn.byteidx(line, cnum)
-    if byte_idx < 0 then
-        -- Cursor is beyond string end, use line length
-        byte_idx = #line
-    end
-
-    local i = byte_idx
-    local preline
-    local iterations = 0
-    while true do
-        preline = line:sub(1, i)
-        local char_cnt = vim.fn.strchars(preline)
-
-        if cnum <= char_cnt or i >= #line then
-            break
-        end
-
-        i = i + 1
-        iterations = iterations + 1
-        -- Safety check: prevent infinite loop (should never hit 100 iterations)
-        if iterations > 100 then
-            vim.notify(
-                "[R.nvim] get_word: Unexpected loop iteration limit reached. " ..
-                "line: '" .. line .. "', cnum: " .. tostring(cnum),
-                vim.log.levels.WARN
-            )
-            break
-        end
-    end
-
+    if byte_idx < 0 then byte_idx = #line end
+    local preline = line:sub(1, byte_idx)
     local pattern = pttrn and pttrn
         or "([%a\192-\244\128-\191_.][%w_.:@$\192-\244\128-\191]*)$"
     local wrd = preline:match(pattern)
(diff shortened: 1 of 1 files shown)


jalvesaq:
Thank you for fixing this! It looks good!

Could you replace

`vim.notify("[R.nvim] ...", vim.log.levels.WARN)`

with

`warn("...")`

where `...` is the warning message? We have a `require("r.log").warn` that includes `R.nvim` as the title of the notification and sets its log level to `WARN`.

jalvesaq:
Do we still need the loop? I don't have emojis completion, but I tried a line with the emoji from your example (📊), and it seems that the loop never runs. So, now that you have found the correct way of slicing the line, would it work if we deleted the loop:

```lua
local get_word = function(line, cnum, pttrn)
    -- Convert character position to byte position for line:sub()
    -- byteidx returns -1 if cnum is beyond the string length
    local byte_idx = vim.fn.byteidx(line, cnum)
    if byte_idx < 0 then
        -- Cursor is beyond string end, use line length
        byte_idx = #line
    end
    local i = byte_idx
    local preline = line:sub(1, i)
    local pattern = pttrn and pttrn
        or "([%a\192-\244\128-\191_.][%w_.:@$\192-\244\128-\191]*)$"
    local wrd = preline:match(pattern)
    return wrd
end
```

jalvesaq:
We don't need the variable `i`; the code can be:

```lua
local get_word = function(line, cnum, pttrn)
    local byte_idx = vim.fn.byteidx(line, cnum)
    if byte_idx < 0 then byte_idx = #line end
    local preline = line:sub(1, byte_idx)
    local pattern = pttrn and pttrn
        or "([%a\192-\244\128-\191_.][%w_.:@$\192-\244\128-\191]*)$"
    local wrd = preline:match(pattern)
    return wrd
end
```

jiangyun-fun:
Thanks for the review! I've simplified the function as you suggested. The `byteidx()` call does indeed give the correct position directly, so the loop was defensive programming that wasn't actually needed.

jalvesaq:
Thank you!

PR #497, opened 2026-01-29
fix: prevent JSON truncation in completionItem/resolve responses

jiangyun-fun:
fix: prevent JSON truncation in completionItem/resolve responses
fix: use robust raw string syntax for Python chunks


jalvesaq:
Thank you!

PR #498, opened 2026-01-31
Adds nix flake based development environment setup

PMassicotte:
I have switched to NixOS a few months ago, and this will use all the needed dev tools for this project.

This is not affecting the plugin, it is only useful for developers using nix flakes.


jalvesaq:
Does it have to be in the R.nvim directory? What if we create a repository for extra code, such as this Nix flake setup, a Debian setup, examples of `~/.config/nvim` directory (minimal config, complete IDE), etc.?

PMassicotte:
Nix flakes must be at the repo root for nix develop to work seamlessly. Moving it elsewhere means developers would need nix develop github:org/extras-repo or clone a second repo, which adds friction. But if you prefer, I can just have these files in my local repo uncommitted. Absolutely no problems :)

jalvesaq:
The files can be here. I'll merge the pull request.

PMassicotte:
> What if we create a repository for extra code, such as this Nix flake setup, a Debian setup, examples of `~/.config/nvim` directory (minimal config, complete IDE), etc.?

I still think this would be a very good idea for Debian/nixos setup with examples.

jalvesaq:
Any suggestion of name for this repository?

PMassicotte:
Something like `r-nvim-config-examples`?

jalvesaq:
> Something like r-nvim-config-examples?

It can be this name. Would you want to start it? I think it would have one directory for each example.

PMassicotte:
I went with https://github.com/R-nvim/R.nvim-config-examples.

> It can be this name. Would you want to start it? I think it would have one directory for each example.

Good idea!

gueyenono:
@PMassicotte Thank you so much for this. I can't wait to see some examples. Although I don't use NixOS, my entire computer configuration is managed by Nix home manager.

PMassicotte:
> @PMassicotte Thank you so much for this. I can't wait to see some examples. Although I don't use NixOS, my entire computer configuration is managed by Nix home manager.

Thank you!

Next week will be mostly a conference week, so I wont have much time, but I will work on that asap.

PR #506, opened 2026-02-23
Fix the command `:RHelp`

jalvesaq:
`:RHelp` now works with symbols such as `%`, `(` and `|`

### DIFF
--- a/lua/r/server.lua
+++ b/lua/r/server.lua
@@ -10,7 +10,7 @@ local o_err = {}
 local rhelp_list = {}
 local lob = {}
 local pkgbuild_attempt = false
-local libs_in_rns = ""
+local new_libs_in_rns = ""
 local building_objls = false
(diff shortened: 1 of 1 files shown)


jalvesaq:
How to try the new `:RHelp` command:

- If you have autocompletion in the command line, just type `:RHelp ` and anything more, such `:RHelp a` or `:RHelp |`.
- If you don't have autocompletion enabled for the command line, press `<c-d>` to trigger the completion.

PMassicotte:
Very good! Some comments coming.

PMassicotte:
Just few comments, you can check if that makes sense, but all good on my side. Good job.

jalvesaq:
I changed the name of the variable. I think the code is clearer now. Thank you for reviewing!

PMassicotte:
yay!

PR #510, opened 2026-03-02
fix(run.lua): replace backslashes in command to prevent escape issues in insert_commented()

PMassicotte:
There was some bad escaping in the last commit (659dc18aba2d).

Use this:

```lua
:lua require('r.run').insert_commented()
```

On this:

```r
# Bad
gsub("\\d+", "X", "abc123def456")
# [1] "abc123def456"
```

With these changes:

```r
# Good
gsub("\\d+", "X", "abc123def456")
# [1] "abcXdefX"
```


jalvesaq:
Thank you!

PR #512, opened 2026-03-03
feat(config.lua, utils.lua): add use_git_files option to improve file search efficiency

PMassicotte:
When working in large projects (>1M data files for example), goto definition is noticeably slow (on the first use) because R.nvim searches all R files to build scope definitions, including files not tracked by git (build artifacts, caches, generated files, etc.).

The new `use_git_files` option skips git ignored files during file search, dramatically improving performance in large git repositories.


jalvesaq:
Thank you!

PR #514, opened 2026-03-11
New option: roxygen_hl

jalvesaq:
Add completion within ROxygen comments if roxygen_hl = true.

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -338,7 +338,8 @@ marks are not visible, but there are plugins that highlight them.
 Sending a selection of lines is actually a special case of sending a marked
 block with the marks being `'<` and `'>`. That is the reason why the command
 to send selection also works in Normal mode: it will send the lines that
-were selected for the last time.
+were selected for the last time. While the cursor is in an ROxygen line,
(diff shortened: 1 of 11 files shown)


PMassicotte:
That is pretty nice!

image

Completion working pretty fine too. Checking code now.

PMassicotte:
I was in the process of testing more, but we got power outage, but I was not able to get completion for "@". I will check that out when power comes back.

jalvesaq:
I'm reviewing Copilot's review...

jalvesaq:
Finished. The four suggestions were correct. I didn't know that printing a NULL string could crash C code on some platforms, and I didn't check the return value of `vim.fn.search()` correctly. The other two suggestions were related to documentation.

jalvesaq:
I squashed the commits and force-pushed the new consolidated commit.

PMassicotte:
I think we can merge then. Excellent job and sending example code to R is just very nice. :1st_place_medal:

jranke:
Many thanks to you - I am very much looking forward to finding the time to test this, but may take a week or two till that happens.

PR #516, opened 2026-03-14
feat(lsp): add targets pipeline symbol recognition for goto-definition and find-references

PMassicotte:
Basically, it is just adding support for `targets` symbols.

Peek 2026-03-14 13-37

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -1515,6 +1515,14 @@ Meaning of options:
 
   - `references`: Enable the references provider
 
+Note: The definition and references providers also recognize symbols defined
+in `{targets}` pipelines. Target-defining calls (e.g., `tar_target()`,
+`tar_file()`) where the first argument is an unquoted identifier are indexed
+as symbol definitions. Consumer calls such as `tar_read()` and `tar_load()`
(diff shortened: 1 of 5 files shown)


PMassicotte:
I made some changes suggested by Copilot. It should be good to go.

jalvesaq:
It's interesting, although I didn't understand well what was happening in the GIF. I didn't know the `targets` library, and after manually typing part of the code from the example and running `:InspectTree`, I didn't see nodes named "target_name". Does it have any extra prerequisites to work?

PMassicotte:
No prerequisites are needed. In `targets`, variables are defined using `tar_*` functions.

If you run the code shown in the GIF with `tar_make()` and then load the results with `targets::tar_load_everything()`, all four variables will be created in the current environment.

Peek 2026-03-15 06-56

This makes the workflow transparent for users, even if they do not use `targets` themselves.

In short, `targets` defines variables using `tar_target(x, 1 + 1)` instead of the usual `x <- 1 + 1`. Variable assignment is captured by treating the first argument in `tar_*(var, ...)` as the variable being assigned.

jalvesaq:
Thanks!

PR #518, opened 2026-03-15
fix source_lines (closes #517)

### LINKED ISSUE #517: Sending a paragraph creates local v:null file and fails to source
**Overview**

Sending a paragraph with \pd fails. Same for selection. The Lua function writing the temporary file seems to be passing a `nil` path, resulting in a literal file named `v:null `being created in the local working directory.

This seems to relate `max_paste_lines` (e.g. reproducible for any line of code if `max_paste_lines  = 0`). 

**Steps to Reproduce:**

1. Use config: `require("r").setup({ max_paste_lines = 0 })`
2. Open an R file and write a multi-line code block/paragraph.
3. Press \pd to send the paragraph to the console.

The following error appears in the R console and a local file `v:null `with the paragraph content is created:
```
Error in file(filename, "r", encoding = encoding) : 
  cannot open the connection
In addition: Warning message:
In file(filename, "r", encoding = encoding) :
  cannot open file '/dev/shm/R.nvim-xtf/Rsource-35753': No such file or directory

```

**Expected Behavior:**
The paragraph is written to the correct temporary file in the RAM disk (e.g., /dev/shm/R.nvim-xxx/Rsource-yyy), and the R console cleanly executes Rnvim.source('/dev/shm/.../Rsource-yyy').

**Environment:**

- Neovim: v0.11.6 (Stable AppImage)
- R.nvim: Latest (lazy update)
- OS: Ubuntu / Linux (24.04)

PMassicotte:
I think that commit 6813c1a renamed source_write/source_read → source_file in config.lua but never updated send.lua.

### DIFF
--- a/lua/r/send.lua
+++ b/lua/r/send.lua
@@ -205,7 +205,7 @@ end
 ---@param what string|nil Additional operation to perform
 ---@return boolean
 M.source_lines = function(lines, what)
-    require("r.edit").add_for_deletion(config.source_write)
+    require("r.edit").add_for_deletion(config.source_file)
 
     local rcmd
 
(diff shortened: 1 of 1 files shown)


jalvesaq:
Thank you, @PMassicotte!

PR #520, opened 2026-03-17
fix(lsp): pass position params and correct buffer to LSP handlers

PMassicotte:
No user facing changes in this PR.

It just fix LSP handlers for definition, references, and implementation to use the position parameters from the LSP request (textDocument/uri, line, character) rather than the live cursor position. This ensures correct behavior when the client sends a request from a position that doesn't match the current cursor.

### DIFF
--- a/lua/r/lsp/utils.lua
+++ b/lua/r/lsp/utils.lua
@@ -421,15 +421,6 @@ function M.deduplicate_locations(locations)
     return unique_locs
 end
 
---- Validate and get keyword at cursor with error handling
----@return string? keyword, string? error
-function M.get_keyword_safe()
-    local ok, word = pcall(require("r.cursor").get_keyword)
-    if not ok then return nil, "Failed to get keyword" end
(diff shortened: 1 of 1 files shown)


jalvesaq:
Thanks!

The function `lsp/utils.get_keyword_safe()` is no longer used and can be deleted.

The function `cursor.get_keyword()` was replaced in some places with `utils.get_word_at_bufpos()`, but is still used in three places. Should it always be replaced with `utils.get_word_at_bufpos()`?

PMassicotte:
Good catch for the unused function.

As for the `cursor.get_keyword()` function, it was not replaced with `utils.get_word_at_bufpos()` in the remaining three call sites (`run.lua`, `lsp/init.lua` hover and signature) because it captures package-qualified names such as `pkg::func`, which treesitter returns as a compound node rather than an `identifier`.

jalvesaq:
Thanks! Merging it now...

PR #521, opened 2026-03-18
Ensure that variables defined in nested functions are correctly resolved

PMassicotte:
I found a bug this morning while working.

With this code:

```r
# Nested function test
outer <- function() {
  x <- 20 # Shadows file-level x

  inner <- function() {
    x <- 30 # Shadows outer x
    print(x) # Should resolve to 30
  }

  print(x) #  ← cursor on x, it should resolve to 20
  inner()
}
```

With the current implementation, it was resolving to `x <- 30` in the `inner()` function. I have added some guards to respect scope boundaries. It should now resolve to `x <- 20` as expected.

I also added more tests with various nesting levels to make sure all is resolved ok.

### DIFF
--- a/lua/r/lsp/ast.lua
+++ b/lua/r/lsp/ast.lua
@@ -61,6 +61,29 @@ function M.find_ancestor(node, node_types)
     return nil
 end
 
+--- Walk up tree to find ancestor of type, stopping at a boundary node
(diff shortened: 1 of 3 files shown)


jalvesaq:
Thank you!

PR #533, opened 2026-03-26
feat: add support for bash code chunks

jiangyun-fun:
- Add bash to supported chunk languages in quarto.lua
- Implement TreeSitter-based code extraction for bash chunks
- Optimize Python chunk sending with TreeSitter-based full expression detection
- Add dedent utility for proper Python code handling
- Update messages and docs to include bash alongside R and Python
- Add test cases for bash chunk detection


PMassicotte:
Some comments:

The Python and Bash branches in `M.line` do basically the same thing — get treesitter code, concat, wrap, send, move cursor. Could pull that into a small helper:

```lua
local function send_chunk_line(chunk, line, lnum, lang, stop_fn, wrap_fn, m)
    local lines
    lines, lnum = get_ts_code_to_send(chunk, line, lnum, lang, stop_fn)
    local code = table.concat(lines, "\n")
    code = wrap_fn(code)
    local ok = M.cmd(code)
    if ok and m == "move" then
        local last_line = vim.api.nvim_buf_line_count(0)
        vim.api.nvim_win_set_cursor(0, { math.min(lnum, last_line), 0 })
        cursor.move_next_line()
    end
end
```

### `dedent` only applied to Python, not bash

This is the case in `codelines_from_chunks`, `source_lines`, and `M.line`. If bash chunks can end up indented (nested Quarto structures etc.), that'll cause problems. If there's a reason bash doesn't need it, worth dropping a comment so the next person doesn't wonder.

### `is_bash` allocates a table every call for one key

```lua
M.is_bash = function(lang)
    local bash_langs = { bash = true }
    return bash_langs[lang] or false
end
```

Maybe could just be `return lang == "bash"`?

### Silent TS fallback

If someone doesn't have the bash (or python) TS grammar installed, `get_ts_code_to_send` quietly falls back to single-line sends. They'll have no idea why multi-line stuff isn't working. A one-time warning would go a long way here.

PMassicotte:
A bit busy today, but will try this PR soon, thank you for your contribution.

jiangyun-fun:
@PMassicotte  Thanks for the thorough review and great suggestions! I’ve gone ahead and pushed
fixes for all four points:

- **Refactoring `M.line`:** Done. I extracted the duplicated logic into a
  `send_chunk_line` helper in `lua/r/send.lua`. It DRYs up the code, and I also
  baked the `utils.dedent()` call directly into the helper before the wrapper
  function is applied.

- **`dedent` for Bash:** Great catch regarding nested Quarto structures, leaving
  the indentation could cause issues with things like heredocs. I've applied
  `utils.dedent()` to Bash blocks in `codelines_from_chunks` and `source_lines`
  (and `M.line` handles it via the new helper mentioned above).

- **`is_bash` optimization:** Fixed. I dropped the table allocation and updated
  it to a straightforward `return lang == "bash"`.

- **Silent TS fallback:** I've implemented a one-time warning system. It uses a
  module-local `ts_warned` table to track the warning state per-language and
  fires a `warn()` if the required TreeSitter grammar is missing. Users will now
  clearly see why multi-line detection is falling back to single-line sends.

PMassicotte:
I will test it later today, but looks cool.

PMassicotte:
I do not have much time to look at this right now, but this is not working on my side. Looks like there is an issue with escaping with the R terminal I am using.

Peek 2026-03-27 13-59

Tomorrow I am free (finally), I will take a look to see what is going on. Your code might be correct, just need to validate few things.

jalvesaq:
Only now have I arrived from a short trip and could try the pull request. It works on my side with this simple Quarto document:

````markdown
---
title: Trying the new chunk support
---

Lorem ipsum dolor sit amet, consectetur
adipiscing elit, sed do eiusmod tempor
incididunt ut labore et dolore magna aliqua.

```{r}
xx <- rnorm(100)
```

Lorem ipsum dolor sit amet, consectetur
adipiscing elit, sed do eiusmod tempor
incididunt ut labore et dolore magna aliqua.

```{python}
xx = 1
yy = 8
while xx < 10:
    xx = xx + 1
    yy = xx + yy
print(yy)
```

Lorem ipsum dolor sit amet, consectetur
adipiscing elit, sed do eiusmod tempor
incididunt ut labore et dolore magna aliqua.

```{bash}
if [ "$HOME" = "nouser" ]
then
    echo "The Home is NO USER"
else
    echo "The HOME is $HOME"
fi
```

Lorem ipsum dolor sit amet, consectetur
adipiscing elit, sed do eiusmod tempor
incididunt ut labore et dolore magna aliqua.
````

- If the cursor is on the chunk header, the whole chunk is sent and the cursor jumps to the next chunk header.
- Sending one line at a time, the while block from the Python code and the if-else-fi block from the Bash code are sent and correctly evaluated.

**Comment by jiangyun-fun** [comment]
> I do not have much time to look at this right now, but this is not working on my side. Looks like there is an issue with escaping with the R terminal I am using.

@PMassicotte I think the issue on `main` may be a pre-existing bug. I tested commit `c37d1cf` (latest on `main`) and it cannot correctly send multi-line R code from Quarto files either. Here's a demo:

https://github.com/user-attachments/assets/97fe7af2-a1e9-4684-9e5f-d57d0355d8f2

I fixed it in my branch by routing R chunks through the same string-parser approach used for Bash and Python.

---

For reference, below is a trace of the root cause (generated with the help of Claude Code, not fully verified since it may be addressed by my latest commit):

**Commit `9dcc072`** ("Fix sending code to R from Rnoweb documents") is where the regression was introduced.

**Before** (`13dc646`):

```lua
local n_lang = vim.o.filetype == "rnoweb" and "r" or nil
local node = vim.treesitter.get_node({
    bufnr = 0,
    pos = { row - 1, col - 1 },
    lang = n_lang,
    ignore_injections = false,
})
```

This used `lang = "r"` for rnoweb files, along with an `ensure_ts_parser_exists()` helper, which told TreeSitter to parse the chunk region with the R parser directly — yielding proper R AST nodes.

**After** (`9dcc072`):

```lua
local node = vim.treesitter.get_node({
    bufnr = 0,
    pos = { row - 1, col - 1 },
    -- Required for quarto/rmd/rnoweb; harmless for r.
    ignore_injections = false,
})
```

It **removed the `lang` parameter** and the `ensure_ts_parser_exists()` function. The fix for rnoweb (#298) relied solely on `ignore_injections = false`, which works for rnoweb but **does not work for quarto/rmd** — `vim.treesitter.get_node()` returns a markdown `code_fence_content` node instead of an R node in Quarto files.

**Summary:**

| Commit    | Change                                                                                     | Effect on Quarto R chunks                                                                    |
| --------- | ------------------------------------------------------------------------------------------ | -------------------------------------------------------------------------------------------- |
| `13dc646` | Original TS approach with `ensure_ts_parser_exists()` + `lang="r"`                         | Worked (explicit R parse of chunk)                                                           |
| `9dcc072` | Removed `ensure_ts_parser_exists()`, removed `lang` param, added `ignore_injections=false` | **Broke** Quarto — `get_node()` returns markdown nodes, walk-up fails, only single line sent |

The root cause: `vim.treesitter.get_node({ ignore_injections = false })` correctly returns R nodes for **rnoweb** (where the injection mechanism works differently), but for **quarto/rmd**, it returns the markdown `code_fence_content` node. The walk-up loop never finds a `program` parent because it's in the markdown tree, so `node` becomes `nil` and only the current line is sent.

PMassicotte:
Alright, on way to work, testing it when I arrive :)

jiangyun-fun:
Hi all, I've pushed a follow-up on top of this PR that makes the chunk language support **config-driven**. Instead of the hardcoded if-else chains for R/Python/Bash, there's now a `chunk_langs` table in config — so adding a new language is just a config entry, no plugin code changes needed.

This means users can do something like:

```lua
require("r").setup({
  chunk_langs = {
    julia = {
      aliases = { "jl" },
      stop_types = { "source_file", "module" },
      dedent = false,
      wrap_inline = function(code)
        return 'JuliaCall::julia_command(r"---(begin\n' .. code .. '\nend)---")'
      end,
      wrap_file = function(filepath)
        return 'JuliaCall::julia_source("' .. filepath .. '")'
      end,
    },
  },
})
```

And Julia chunks in `.qmd` files just work — send line, send chunk, send selection, all of it.

The existing R/Python/bash behavior is unchanged — the defaults reproduce the exact same wrapping and stop types. Docs are updated too.

Would appreciate a review when you all have time. @PMassicotte @jalvesaq

PMassicotte (changes requested):
1. All chunk languages (including R) now go through `send_chunk_line()` → `get_ts_code_to_send().` Previously R used paren-matching (`get_r_code_to_send()`), which doesn't need a TreeSitter parser. The TS fallback only sends the single current line, so multi-line R expressions would break for users without the R TS grammar. R chunks should still use the paren-matching path. Are we ok with users having to get TS for `R.nvim`?

2. Stale \_alias_map cache (quarto.lua:304) M.\_alias_map is built once and never invalidated. If chunk_langs changes after setup, the map is stale. Either rebuild it on each call (it's tiny) or invalidate on config change.

jiangyun-fun:
> 1. All chunk languages (including R) now go through `send_chunk_line()` → `get_ts_code_to_send().` Previously R used paren-matching (`get_r_code_to_send()`), which doesn't need a TreeSitter parser. The TS fallback only sends the single current line, so multi-line R expressions would break for users without the R TS grammar. R chunks should still use the paren-matching path. Are we ok with users having to get TS for `R.nvim`?
> 2. Stale _alias_map cache (quarto.lua:304) M._alias_map is built once and never invalidated. If chunk_langs changes after setup, the map is stale. Either rebuild it on each call (it's tiny) or invalidate on config change.

1. R chunks now requiring TreeSitter

The claim that `get_r_code_to_send()` uses paren-matching without TreeSitter is not quite right. That function actually does use TreeSitter in `send.lua`:

```lua
local node = vim.treesitter.get_node({
    bufnr = 0,
    pos = { row - 1, col - 1 },
    ignore_injections = false,
})
```

The paren-matching path (`paren_diff`) is only used for the `rhelp` filetype (via `get_rhelp_code_to_send`). So both the old and new paths already require the R TreeSitter grammar. The difference is just the API used:

- `get_r_code_to_send()` → `vim.treesitter.get_node()` (buffer parser, for plain `.R` files)
- `get_ts_code_to_send()` → `vim.treesitter.get_string_parser()` (string parser, for chunks in Quarto/Rmd)

Additionally, R.nvim's `check_health` already validates that the R TreeSitter grammar is installed. So any R.nvim user should already have it.

2. Stale `_alias_map` cache

I've removed the cached alias map entirely. `resolve_lang()` now rebuilds it on every call. The table is tiny (3 language entries + their aliases), so the cost is negligible.

PMassicotte:
LGTM, thank you for your contribution. @jalvesaq ?

jalvesaq:
I'll merge it. Thank you, @jiangyun-fun and @PMassicotte!

PR #542, opened 2026-04-03
feat(R.nvim): add support for workspace symbol provider

PMassicotte:
Forgot to add the option to op in/out of the workspace symbols as we did for the other implemented LSP features.


jalvesaq:
I also forgot...

The option is named `workspace_symbols` while the LSP feature is `WorkspaceSymbol`, with "symbol" in the singular form. Shouldn't it be `workspace_symbol` as we already have `document_symbol`?

jalvesaq:
Thank you!

PR #544, opened 2026-04-06
Adding lsp rename capability

PMassicotte:
This is implementing lsp rename into `r_ls`.

Here is a quick demo (I am using `lspsaga` for my rename UI):

Peek 2026-04-06 07-38

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -1549,6 +1549,7 @@ features in your R.nvim config. Below are the default values:
     document_highlight = true,  -- enable the document highlight provider
     document_symbol = true,     -- enable the document symbol provider
     workspace_symbol = true,    -- enable the workspace symbol provider
+    rename = true,              -- enable the rename provider
     doc_width = 0,
     fun_data_1 = { "select", "rename", "mutate", "filter" },
     fun_data_2 = { ggplot = { "aes" }, with = { "*" } },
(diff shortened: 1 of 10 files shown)


jalvesaq:
Thank you! I know that it's still a draft, and bugs are expected...

I tried to replace `info` in this code from `nvimcom/R/bol.R`:

```r
    if (length(res) == 0 || (length(res) == 1 && res == "")) {
        res <- ""
    } else {
        if (is.null(pkg)) {
            info <- pkgname
            if (!is.na(funcmeth)) {
                if (info != "") {
                    info <- paste0(info, ", ")
                }
                info <- paste0(info, "function:", funcmeth, "()")
            }
            # TODO: Add the method name to the completion menu if (info != "")
        }
    }
```

With the cursor over the first occurrence of `info`, it asks twice for the new name, and only the second works. It renames the first three occurrences of `info`, but not the other three. Maybe this is correct... But then, I undid the changes and repeated the renaming procedure. This time, only the first `info` under the cursor was renamed.

PMassicotte:
Thank you very much for sharing your experience. So here, all 6 instances should be renamed?d

PMassicotte:
I replicated the R code into a lua function to see what lua-lsp would change 

image

We see all the 6 instances highlighted.

With the code change I am about to push, we also have the same behaviour

image

PMassicotte:
I have not found any bugs lately. I think this is ready for review :)

jalvesaq:
The behavior didn't change on my side: I have to enter the new name twice, and only the first three `info` are replaced.

jalvesaq:
I'm sorry, it was my fault. The `languageserver` was enabled. Everything is OK.

PR #548, opened 2026-04-10
Fix: RSendMotion support for motions 0, fFtT, ip/ap, Fixes #547

### LINKED ISSUE #547: (RSendMotion does not support text objects (ip, ap), two-character motions (ftFT, etc.), or motion going to start of line (0)
# Description

`<localleader>m`, i.e. `RSendMotion` followed by a text object like ip or ap does not send the expected region. 
Instead, the prefix (i or a) is silently consumed as the motion, and the second character is used as a normal mode command, e.g. for \mip, p pastes the previous yank.
Additionally, \m0 (send to beginning of line) does not work because 0 is incorrectly read by the count-reading loop.

Both issues come from the `M.motion_wrapper()` implementation in `lua/r/send.lua`.

This can be replicated by trying to use \mip or \map in any context when R.nvim is running.

I am opening this issue to reference in the PR

PeterPavicic:
Additionally to the fix for the motions, the pre-commit stylua hook changes the format of send_chunk_line, I'm not sure if this is intentional.

### DIFF
--- a/lua/r/send.lua
+++ b/lua/r/send.lua
@@ -238,7 +238,16 @@ local M = {}
 ---@param should_dedent boolean Whether to dedent the code
 ---@param m string|nil Movement mode ("move" or nil)
 ---@return boolean Whether the command was sent successfully
-local function send_chunk_line(chunk, line, lnum, lang, stop_fn, wrap_fn, should_dedent, m)
+local function send_chunk_line(
+    chunk,
+    line,
+    lnum,
+    lang,
+    stop_fn,
+    wrap_fn,
+    should_dedent,
+    m
+)
     local lines
     lines, lnum = get_ts_code_to_send(chunk, line, lnum, lang, stop_fn)
     local code = table.concat(lines, "\n")
@@ -507,8 +516,24 @@ M.motion_wrapper = function()
 
     motion = char
 
-    -- If first char is a digit, keep reading digits
-    while char:match("%d") do
+    -- If first char is a non-zero digit, keep reading digits
+    if char:match("[1-9]") then
+        repeat
+            ok, char = pcall(vim.fn.getcharstr)
+            if not ok or char == "\27" then return end
+            motion = motion .. char
+        until not char:match("%d")
+    end
+
+    -- If part of two-character motion is typed (in/around), keep reading
+    if
+        char == "i"
+        or char == "a"
+        or char == "f"
+        or char == "F"
+        or char == "t"
+        or char == "T"
+    then
         ok, char = pcall(vim.fn.getcharstr)
         if not ok or char == "\27" then return end
         motion = motion .. char


mschilli87:
fixes https://github.com/R-nvim/R.nvim/issues/547

---
@PeterPavicic: Thank you.
AFAIK, you need to mention the issue in the commit message (and/)or PR text to have GitHub automatically cross-reference issue and PR and have mergin the latter auto-close the former.
I hope that this comment is sufficient to trigger this, else you might have to edit your message (and/)or force-push an amended commit with an adjusted message.

PeterPavicic:
Thank you, it worked this time! Seems that the PR text wasn't sufficient, but mentioning in the commit worked

jalvesaq:

I tried the pull request with the following code:

```r
if (TRUE) {
    x <- rnorm(100)
    y <- rnorm(100)
}
```

If the cursor is anywhere between the curly brackets, and I type `<LocalLeader>mi{`, the two `rnorm` lines are sent, as expected. This fixes a bug from the main branch!

> Additionally to the fix for the motions, the pre-commit stylua hook changes the format of send_chunk_line, I'm not sure if this is intentional.

The indentation was not following our conventions. It was fixed in your pull request.

Let's wait for @PMassicotte review...

PMassicotte:
Hi and thank you for your contribution.

I think there is an issue with the change from `char:match("%d")` to `char:match("[1-9]")` that can break counts that contain 0 after a leading non-zero digit (e.g. 10j, 20k, 100j).

Trace through 10j with the new code:

1. Read 1 → char="1", motion="1"
2. While "1" matches [1-9]: read 0 → char="0", motion="10"
3. While "0" doesn't match [1-9]: exit loop, j is never read
4. "0" is not i/a/f/F/t/T → falls through
5. Executes g@10, j is lost and never consumed

I think this would be better:

```lua
if char:match("[1-9]") then
    repeat
        ok, char = pcall(vim.fn.getcharstr)
        if not ok or char == "\27" then
            return
        end
        motion = motion .. char
    until not char:match("%d")
end
```

lua/r/send.lua
```diff
@@ -507,8 +516,22 @@ M.motion_wrapper = function()
...
     motion = char
 
-    -- If first char is a digit, keep reading digits
-    while char:match("%d") do
+    -- If first char is a non-zero digit, keep reading digits
+    while char:match("[1-9]") do
+        ok, char = pcall(vim.fn.getcharstr)
+        if not ok or char == "\27" then return end
+        motion = motion .. char
+    end
```
PMassicotte:
```suggestion
if char:match("[1-9]") then
    repeat
        ok, char = pcall(vim.fn.getcharstr)
        if not ok or char == "\27" then
            return
        end
        motion = motion .. char
    until not char:match("%d")
end
```

PMassicotte: I have not had time to test it, so check before :)

PMassicotte (changes requested):

PeterPavicic:
I added the change written by @PMassicotte , good catch on the bug!
I hope force-pushing like this is fine, I don't have a lot of experience with open source contribution yet, to be honest.
Please let me know if a different solution should be implemented so you are also credited in the commit

Two more things I have noticed while editing/testing are that:

- Movement downwards (with j) does not move the cursor while movement upwards (with k) does, except from the last line
- It would be cool to have characterwise motions send only the parts between the movements (i.e. \mfa reads the motion, and only executes until `a`, currently this sends the whole lines

Something more for me to potentially work on, I'll see what I have time for in the future

PMassicotte:
LGTM, @jalvesaq ?

jalvesaq:
Thank you, @PeterPavicic and @PMassicotte !
I'm on a trip now, and would be able to look at this only on Monday. Please, feel free to merge it!

PMassicotte:
I have not found issues, so I will merge.

PR #550, opened 2026-04-15
fix(send.lua): handle stop nodes correctly by stepping into first child to ensure proper node traversal

PMassicotte:
Thank you @warneford for reporting the issue and providing a starting point to solve it.


jalvesaq:
Thank you! Merging it...

PR #552, opened 2026-04-21
fix(nvimcom): fix single-column data.frame crash in nvim_viewobj

### LINKED ISSUE #551: `\rv` crashes on single-column data frames
### Config

```lua
view_df = {
  save_fun = "function(obj, obj_name) {f <- paste0('/tmp/', obj_name, '.parquet'); arrow::write_parquet(obj, f) ; f}",
},
```

### Steps to reproduce

1. Create a single-column data frame with more than `ceiling(10000 / ncol)` rows
2. Place cursor on `df` and press `\rv`

```r
df <-
  base::data.frame(
    col1 = 1:100000
  )
```

Also reproduces with plain `tidytable` or `data.table` — any single-column object.

### Expected behavior

The object is subsetted to `ceiling(10000 / ncol)` rows and passed to `save_fun`.

### Actual behavior

```
Error in `tryCatchList()`:
! Object must be coercible to an Arrow Table using `as_arrow_table()`
Caused by error in `as_arrow_table()`:
! No method for `as_arrow_table()` for object of class integer
```

### Root cause

`nvim_viewobj` subsets with `o <- o[1:nrows, ]` without specifying `drop`. The default value of `drop` in `[.data.frame` is:

```r
drop = if (missing(i)) TRUE else length(cols) == 1
```

When there is only one column, this evaluates to `drop = TRUE`, which silently converts the data frame to a bare vector. The vector then crashes when `save_fun` passes it to `arrow::write_parquet`.

Two-column (and wider) data frames are unaffected because `length(cols) > 1` makes `drop = FALSE`.

jiangyun-fun:
`[.data.frame` has `drop = if (missing(i)) TRUE else length(cols) == 1` as default, not `drop = FALSE`. This means single-column data.frames are silently dropped to bare vectors when subsetted as `o[1:nrows, ]`.

Fix by explicitly passing `drop = FALSE`.
Fixes #551

### DIFF
--- a/nvimcom/R/misc.R
+++ b/nvimcom/R/misc.R
@@ -114,7 +114,7 @@ nvim_viewobj <- function(
             nrows <- ceiling(10000 / ncol(o))
         }
         if (nrows != 0 && nrows < nrow(o)) {
-            o <- o[1:nrows, ]
+            o <- o[seq_len(nrows), , drop = FALSE]
         }
         if (!is.null(R_df_viewer)) {
             cmd <- gsub("'", "\x13", R_df_viewer)


PMassicotte:
LGTM. Thank you for your contribution.

PR #554, opened 2026-04-22
feat(lsp): add pipe chain column resolution for tidyverse data-masking

PMassicotte:
Introduce a new module `pipe.lua` to handle tidyverse data-masking semantics in R.nvim LSP. This allows for resolving column definitions in pipe chains, enhancing the goto definition and document highlight features.

Peek 2026-04-22 14-38


jalvesaq:
I confirm the bug and that the fix works! Thank you!

PR #555, opened 2026-04-24
feat(r.lua): fix pipe indentation within R functions

PMassicotte:
This PR makes sure that new lines are correctly indented within R functions for both pipe `|>` and `+` end lines.

Peek 2026-04-24 08-36

### DIFF
--- a/after/ftplugin/r.lua
+++ b/after/ftplugin/r.lua
@@ -1,0 +1,9 @@
+-- Override indentexpr to fix trailing-pipe/ggplot2 + indentation.
+-- vim.schedule defers until after nvim-treesitter's FileType autocmd runs,
+-- so our setting wins the race.
+vim.schedule(function()
+    if vim.bo.filetype == "r" then
+        vim.bo.indentexpr = "v:lua.require'r.indent'.get_indent(v:lnum)"
+    end
+end)
+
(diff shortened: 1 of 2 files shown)


jalvesaq:
I can't see any bugs on the main branch. This code is correctly indented:

```r
ffff <- function() {
    dfrm |>
        mutate() |>
        abcdef()

    ggplot() +
        ghijk() +
        lmnop()
}
```

PMassicotte:
I mean, when you enter a new line, it places it correctly on your side (independently how you typed `|>|` for example)?

jalvesaq:
Yes, whether I manually type `|>` or use the shortcut `<LocalLeader>,`, when I open a new line, it's automatically indented and the cursor placed below the first character of the previous line.

When I save the file, and the `conform` plugin runs `air`, nothing changes because the indentation is already correct.

PMassicotte:
oh wow it never worked for me. Let me check that. Not sure what I have wrong on my side.

PMassicotte:
I added that to my lua config and not it works.

```lua
local ft = vim.bo.filetype
if ft ~= "r" and ft ~= "rnoweb" then
    vim.bo.indentexpr = "v:lua.require'nvim-treesitter'.indentexpr()"
end
```

Sorry for the noise!

PR #556, opened 2026-04-24
optionally suppress "building autocompletion for <package>" messages

urtzienriquez:
This PR introduces the option to suppress messages about building autocompletion for packages on startup.

In my case, I have some messages fired with autocmds when I e.g. load a neovim session. Thus,  I would like to suppress those other messages that get on the way.

The behaviour can be controlled with an option in .Rprofile

options(nvimcom.build_messages = FALSE) # supresses messages

If not set, or set to TRUE, messages are shown

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -473,6 +473,13 @@ trigger the completion after the double colon).
 For both `library()` and `require()`, when completing the first argument, the
 popup list shows the names of installed packages.
 
+At startup, R.nvim emits messages about the packages for which completion 
+lists are built. To suppress these messages, add the following to your 
+~/.Rprofile:
+
+>r
+   options(nvimcom.build_messages = FALSE)
+<
 
 ------------------------------------------------------------------------------
 4.3.1. Quarto YAML completion                          *R.nvim-quarto-completion*
--- a/lua/r/server.lua
+++ b/lua/r/server.lua
@@ -193,7 +193,6 @@ local init_exit = function(_, data, _)
 end
 
 local build_objls_exit = function()
-    vim.schedule(function() vim.api.nvim_echo({ { " " } }, false, {}) end)
     edit.add_to_debug_info(
         "stderr of last completion data building",
         table.concat(o_err, "\n")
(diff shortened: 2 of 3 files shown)


jalvesaq:
Thanks, I agree. Please, try:

```r
    if (isTRUE(getOption("nvimcom.build_messages", TRUE))) {
         msg <- paste0("ECHO: Building completion list for \"", p, "\"\x14\n")
         cat(msg)
         flush(stdout())
    }
```

urtzienriquez:
great! I have also seen that there was an empty message being emitted in the `build_objls_exit` function in server.lua

Removing that completely silences messages of loading packages, and when not silenced, it ends the message cleanly (without an additional empty line). Was that intentional?

jalvesaq:
> great! I have also seen that there was an empty message being emitted in the build_objls_exit function in server.lua

Yes, that is intentional. It clears the status line. However, if no message was ever printed, there is nothing to clear. One solution is to create a local variable, such as `local has_msg = false`, and set its value to `true` when a message is printed. Then, the empty line would be printed only if `has_msg` were `true`.

urtzienriquez:
Ah, I see! The empty message is fine then but it is only useless if using ui2 and "msg" as the target (that was the empty message left that I was getting). I have added a condition to only print the empty message if ui2 is not enabled, configured, or if the target is not "msg" when it is configured. 

Does it make sense?

PMassicotte:

I think this should work with only the R change you made no? I will print a single space `{ " " }`? Also maybe add documentation about that in `nvim.txt`.

jalvesaq:
@urtzienriquez, I'm sorry if I wasn't clear.
I changed the code pull request. Now, we use a local variable as I was trying to suggest.

urtzienriquez:
Oh! that is much cleaner! Thanks for the fix! :)

I have added a line in the docs to document this option. Don't know if it is in the most appropriate section though

jalvesaq:
Thank you, @urtzienriquez and @PMassicotte! I'll merge it now.

urtzienriquez:
Thank you @jalvesaq and @PMassicotte !!

PMassicotte:
Thank you both!

PR #559, opened 2026-04-26
fix: make nvim-treesitter optional (fixes #558)

### LINKED ISSUE #558: Warning for missing parsers is always shown when nvim-treesitter is not used
Hi there, I've noticed a potential issue / inconvenience with R.nvim when you are not using nvim-treesitter to install and manage your tree-sitter parsers.

As soon as you open any *.r file you are greeted with this warning message:
```
R.nvim warning:
  R.nvim requires nvim-treesitter. Please install it and the parsers for "r", "markdown", "rnoweb", and "yaml".
```

After dismissing the message R.nvim can be started and works as expected.

Due to the archival of nvim-treesitter I have installed the parsers through another plugin (https://github.com/romus204/tree-sitter-manager.nvim). All the necessary parsers are available on my system (see my :checkhealth below):

```
Checking applications and plugins: ~
- ✅ OK Neovim version: 0.12.2+v0.12.2
- ✅ OK C compiler (`gcc` or `clang`) found.
- ✅ OK Tree-sitter command line application (`tree-sitter`) found.
- ⚠️ WARNING `nvim-treesitter/nvim-treesitter` not found. Ignore this if you manually installed the parsers.
- ✅ OK `yaml-language-server` found.

Checking tree-sitter parsers: ~
- ✅ OK `r`  found.
- ✅ OK `markdown`  found.
- ✅ OK `markdown_inline`  found.
- ✅ OK `rnoweb`  found.
- ✅ OK `latex`  found.
- ✅ OK `yaml`  found.
```

Because checkhealth already says, that the warning can be ignored, it might be unnecessary to show the warning to the user when all parsers are installed. In the configuration options I have not found any option to hide the warning above.

PMassicotte:
- Update config, health checks, and documentation to reflect that nvim-treesitter is no longer a required dependency.
- For now, we keep nvim-treesitter for internal tests. We might change that later with the adopted successor to nvim-treesitter.

### DIFF
--- a/lua/r/config.lua
+++ b/lua/r/config.lua
@@ -1309,27 +1309,20 @@ M.check_health = function()
         end
         return false
     end
-    local has_treesitter, _ = pcall(require, "nvim-treesitter")
-    if not has_treesitter then
+
+    local parsers = vim.api.nvim_get_runtime_file(
+        "parser" .. (config.is_windows and "\\" or "/") .. "*.*",
+        true
+    )
+    if
+        not has_parser("r", parsers)
+        or not has_parser("markdown", parsers)
+        or not has_parser("rnoweb", parsers)
+        or not has_parser("yaml", parsers)
+    then
         swarn(
-            'R.nvim requires nvim-treesitter. Please install it and the parsers for "r", "markdown", "rnoweb", and "yaml".'
+            'R.nvim requires treesitter parsers for "r", "markdown", "rnoweb", and "yaml". Please, install them.'
         )
-    else
-        -- Check if required treesitter parsers are available
-        local parsers = vim.api.nvim_get_runtime_file(
-            "parser" .. (config.is_windows and "\\" or "/") .. "*.*",
-            true
-        )
-        if
-            not has_parser("r", parsers)
-            or not has_parser("markdown", parsers)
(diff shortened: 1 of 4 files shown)


jalvesaq:
Thank you!

lua/r/config.lua
```diff
@@ -1309,27 +1309,20 @@ M.check_health = function()
...
     end
-    local has_treesitter, _ = pcall(require, "nvim-treesitter")
-    if not has_treesitter then
+
+    local parsers = vim.api.nvim_get_runtime_file(
+        "parser" .. (config.is_windows and "\\" or "/") .. "*.*",
+        true
+    )
+    if
+        not has_parser("r", parsers)
```
PMassicotte, line 1318: @jalvesaq In the README, we also ask for csv, latex parsers. Should we add them here too?

jalvesaq: CSV and TSV parsers will be useful only if viewing data.frames as CSV or TSV files (`<LocalLeader>rv`). I'm not sure if we still need the `latex` parser. Anyway, it would be useful only for those editing Rnoweb files. I think it's OK to recommend them in the README, but not annoy those who didn't install them with a warning.

PR #561, opened 2026-04-27
fix(nvimcom.c): ensure compatibility with R version 4.4.0 and above

PMassicotte:
Add conditional compilation to use `Rf_allocLang` for R 4.4.0+. Maintain backward compatibility with older R versions using `Rf_allocList`.


jalvesaq:
I pushed two commits, but didn't change the required R version...

PMassicotte:
This is all good on my side!

jalvesaq:
Let's wait for @ClemensKohl's confirmation that this pull request fixes the bug...

ClemensKohl:
I just tried it out, and it installed without a problem! Thanks a lot for looking into it!

PR #564, opened 2026-04-29
feat(quarto): add rnoweb chunk scanning

jiangyun-fun:
- Add get_rnw_code_chunks() to scan <<...>>= / @ markers and produce
  Chunk objects, enabling the send_chunk_line TreeSitter path for
  rnoweb files
- Add parse_rnw_params() for rnoweb chunk header parameters
- Dispatch chunk lookup to get_rnw_code_chunks for rnoweb filetype
  in get_current_code_chunk, get_chunks_above_cursor, and
  get_chunks_below_cursor
- Check if cursor is already on a chunk boundary (@ or ```) before
  scanning forward, so that landing on @ after sending the last code
  line correctly triggers next_chunk navigation.
- Require non-empty chunk for send_chunk_line path to prevent
  nil error when chunk:get_content() is called on an empty table
- Show "Not inside an R code chunk." for rnoweb/markdown/rmd/quarto
  files when cursor is outside code sections

### DIFF
--- a/lua/r/quarto.lua
+++ b/lua/r/quarto.lua
@@ -174,6 +174,66 @@ M.get_code_chunks = function(bufnr)
     return code_chunks
 end
 
+--- Get code chunks from an Rnoweb (.Rnw) document by scanning <<...>>= / @ markers.
+--- Avoids the markdown TreeSitter parser which does not understand Rnoweb syntax.
+---@param bufnr integer The buffer number.
+---@return table
+M.get_rnw_code_chunks = function(bufnr)
+    bufnr = bufnr or vim.api.nvim_get_current_buf()
+    local lines = vim.api.nvim_buf_get_lines(bufnr, 0, -1, false)
+    local chunks = {}
+
+    local i = 1
+    while i <= #lines do
+        local params_str = lines[i]:match("^<<(.*)>>=%s*$")
+
+        if params_str then
+            local start_row = i
+
+            -- Find the closing @ (must start at column 0)
+            local end_row = nil
+            local j = i + 1
+            while j <= #lines do
+                if lines[j]:find("^@%s*$") then
+                    end_row = j
+                    break
+                end
+                j = j + 1
+            end
+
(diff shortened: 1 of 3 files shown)


jalvesaq:
Thank you, @jiangyun-fun! The code looks good. I tried it quickly, and it worked.

My only observation is that it's strange to me to see `rnw`  function in a Lua script called `quarto.lua`. Maybe we should rename `lua/r/quarto.lua` as `lua/r/chunk.lua`.

lua/r/quarto.lua
```diff
@@ -230,7 +323,12 @@ M.get_current_code_chunk = function(bufnr)
     if not cursor then return {} end
     local row, _ = unpack(cursor)
 
-    local chunks = M.get_code_chunks(bufnr)
+    local chunks
+    if vim.bo.filetype == "rnoweb" then
+        chunks = M.get_rnw_code_chunks(bufnr)
+    else
+        chunks = M.get_code_chunks(bufnr)
+    end
```
PMassicotte, line 331:
```suggestion
  local function get_chunks_for_buffer(bufnr)                                                                                                                                                                                    
      if vim.bo.filetype == "rnoweb" then                                                                                                                                                                                        
          return M.get_rnw_code_chunks(bufnr)                                                                                                                                                                                    
      end                                                                                                                                                                                                                        
      return M.get_code_chunks(bufnr)                                                                                                                                                                                            
  end 
  
  local chunks = get_chunks_for_buffer(bufnr)
```

The code above is used at 3 places, lines 326, 350 and 378. Maybe extract into a small helper.

jalvesaq:
This pull request is for the "improve_rnw" branch, not for the "main" branch. Perhaps we could merge it, @PMassicotte could create the helper functions, then we would wait for @jiangyun-fun to review the changes later since he might be in a different time zone. What do you think, @PMassicotte ?

PMassicotte:
All good with me!

PR #562, opened 2026-04-28
Improve highlighting of Quarto and Rnoweb blocks

jalvesaq:

PR #565, opened 2026-04-29
Fix support for Rnoweb

jalvesaq:


jalvesaq:
Remaining problems:

- The name `quarto.lua` is misleading because the code gets and highlights blocks of R code from Rmd, Quarto, and Rnoweb. However, when I renamed it as `chunk.lua`, git lost its history (see #562).
- The option `quarto_chunk_hl` also has a misleading name because it also defines the highlighting of chunks of R code in Rnoweb files. I think we should rename it as `chunk_hl`.

jiangyun-fun:
Hi @jalvesaq,

Sorry for the late reply. I've been busy with work lately and haven't been able to keep up with the PRs.

I tried to address the two remaining issues:

1. Renamed `quarto.lua` → `chunk.lua`.
2. Renamed `quarto_chunk_hl` → `chunk_hl`. Updated everywhere including docs. Also added a migration path so existing
   configs with the old name still work (with a deprecation warning).

Details are in #568. I'd appreciate it if you could take a look when you have time.

I should have more free time in the coming days, so I'll be able to respond and iterate quickly on any feedback.

Thanks!

PMassicotte:
I hope I did not mess up

jalvesaq:
The "Migration" code in `config.lua` is in two places. I think it's better to keep the code from lines 1140-1150, before merging `user_opts` and `config`.

jalvesaq:
I added a superfluous check that can be deleted (`user_opts` will always exists):

```diff
diff --git a/lua/r/config.lua b/lua/r/config.lua
index ab39dd8..c77c68b 100644
--- a/lua/r/config.lua
+++ b/lua/r/config.lua
@@ -1139,8 +1139,7 @@ local global_setup = function()
 
     -- Migration: quarto_chunk_hl renamed to chunk_hl
     if
-        user_opts
-        and user_opts.quarto_chunk_hl
+        user_opts.quarto_chunk_hl
         and type(user_opts.quarto_chunk_hl) == "table"
         and user_opts.chunk_hl == nil
     then
```

PMassicotte:
I agree, can yyou make it?

jalvesaq:
Yes.

PMassicotte:
Besides that, this lgtm. Good job!

PMassicotte:
Ready to merge?

jalvesaq:
>  Ready to merge?

Yes, I don't see any problem.

PR #568, opened 2026-05-02
refactor: rename quarto.lua to chunk.lua and quarto_chunk_hl to chunk_hl

jiangyun-fun:
## Summary

- Rename `quarto.lua` → `chunk.lua` via `git mv` (history preserved with `git log --follow`)
- Rename config option `quarto_chunk_hl` → `chunk_hl` (the module handles Rmd, Quarto, **and** Rnoweb, not just Quarto)
- Add backward-compatible migration: users with `quarto_chunk_hl` in their config get values merged into `chunk_hl` with a deprecation warning
- Update all `require("r.quarto")` → `require("r.chunk")` references across the codebase
- Update documentation

## Commits

1. `refactor: rename quarto.lua to chunk.lua` — pure `git mv`, no content changes
2. `refactor: update require("r.quarto") to require("r.chunk")` — all module references
3. `refactor: rename quarto_chunk_hl config option to chunk_hl` — config + docs
4. `fix: add backward-compatible migration for quarto_chunk_hl → chunk_hl` — deprecation path


jalvesaq:
Thank you, @jiangyun-fun! Git blame shows the correct history. After the PR is merged, we can try a minor refactor and create a `r/quarto.lua` with only one function, `command()`. It was when I created this new `quarto.lua` that Git became confused.

I would like to know @PMassicotte's opinion on renaming the option `quarto_chunk_hl` as `chunk_hl`...

lua/r/rmd.lua
```diff
@@ -3,7 +3,7 @@ local config = require("r.config").get_config()
 local get_lang = require("r.utils").get_lang
 local uv = vim.uv
 local chunk_key = nil
-local quarto = require("r.quarto")
+local quarto = require("r.chunk")
```
PMassicotte, line 6: If we rename, we should use
```suggestion
local chunk = require("r.chunk")
```

See down below about quarto references still present.

lua/r/config.lua
```diff
@@ -331,8 +331,8 @@ local hooks = require("r.hooks")
 ---Do `:help quarto_render_args` for more information.
 ---@field quarto_render_args? string
 ---
----How to highlight code blocks in Quarto and Rmd documents.
----@field quarto_chunk_hl? { highlight: boolean, yaml_hl: boolean, virtual_title: boolean, bg: string, events: string }
+---How to highlight code blocks in Quarto, Rmd, and Rnoweb documents.
+---@field chunk_hl? { highlight: boolean, yaml_hl: boolean, virtual_title: boolean, bg: string, events: string }
```
PMassicotte, line 335:
```suggestion
 ---@field chunk_hl? { highlight: boolean, yaml_hl: boolean, virtual_title: boolean, bg: string, events: string }                                                                                                  
 ---                                                                                                                                                                                                               
 ---@deprecated Use `chunk_hl` instead.                                                                                                                                                                            
 ---@field quarto_chunk_hl? { highlight: boolean, yaml_hl: boolean, virtual_title: boolean, bg: string, events: string } 
```

Maybe for the lsp?

PMassicotte (commented):
Here are some comments.

- Several files (rmd.lua, send.lua, utils.lua) still alias require("r.chunk") to a local variable named quarto, which contradicts the rename intent.
- Maybe keep a `quarto.lua` file with this content to avoid breaking changes

```lua
return require("r.chunk")
```

PMassicotte:
> I would like to know @PMassicotte's opinion on renaming the option `quarto_chunk_hl` as `chunk_hl`...

`chunk_hl` is kinda generic. Without context, it is unclear what kind of chunks it refers to. `quarto_chunk_hl` at least hinted at "code chunks in document formats." Maybe better to have `code_chunk_hl`?

jalvesaq:
> Maybe keep a quarto.lua file with this content to avoid breaking changes

The problem is that when I created a `quarto.lua` with any content from the now `chunk.lua`, Git gets confused, and the history of `chunk.lua` gets lost. The new `quarto.lua` inherits the history of the old `quarto.lua`, and `git blame` on `chunk.lua` shows it was created entirely in a single commit.

PMassicotte:
> > Maybe keep a quarto.lua file with this content to avoid breaking changes
> 
> The problem is that when I created a `quarto.lua` with any content from the now `chunk.lua`, Git gets confused, and the history of `chunk.lua` gets lost. The new `quarto.lua` inherits the history of the old `quarto.lua`, and `git blame` on `chunk.lua` shows it was created entirely in a single commit.

Valid point, maybe we can do it in another PR...

jalvesaq:
> vim.notify("[R.nvim] require('r.quarto') is deprecated, use require('r.chunk')", vim.log.levels.WARN)
return require("r.chunk")

OK. This should work since it doesn't include any contents of the old `quarto.lua`. However, I don't see the need for this warning, since normal users will not write Lua code that requires `quarto.lua`.

jalvesaq:
> Valid point, maybe we can do it in another PR...

This might work. We can try after this PR and its parent PR are merged.

PMassicotte:
So, accept like this and polish later with some of my suggestions? I can work on that after the lunch.

jalvesaq:
> So, accept like this and polish later with some of my suggestions? I can work on that after the lunch.

OK. I'm merging this now. Then, we can continue on #565

PMassicotte:
Should I commit some changes about the `quarto` naming consistency in #565? Or start another PR from your branch

jalvesaq:
> Should I commit some changes about the quarto naming consistency in https://github.com/R-nvim/R.nvim/pull/565? Or start another PR from your branch

On 565 is better. The only thing that we cannot do (yet) is to resurrect `quarto.lua`. I'm updating the documentation, isolating `chunk_hl` in a new section.

PR #569, opened 2026-05-02
Move Quarto specific code to quarto.lua

jalvesaq:


jalvesaq:
This pull request does a small refactoring. Move one function that is Quarto-specific to `r/lua/quarto.lua`.

The Selene error is caused by the branch name `"add_quarto.lua"`, which Selene sees as a Lua file in the `.git/refs/origin/` directory.

jalvesaq:
It seems that the `args` of the selene action should be:

```
 args: --config selene.toml lua
```

PMassicotte:
LGTM, I will merge and see what is going on with `selene`.

PMassicotte:
All looks good, thank you very much!

PR #570, opened 2026-05-03
The remote access only works if using Neovim's built-in terminal

jalvesaq:


jalvesaq:
I discovered that remote access via SSH only works with Neovim's built-in terminal emulator. This pull request adds a note on this limitation and is ready for review (unless someone wants to implement support for running R through SSH in an external terminal).

PMassicotte:
Good to document it. I might start working more over SSH soon, so it will be a good opportunity to implement that in another terminal. Thank you!

PR #575, opened 2026-05-10
Support for RTypst

jalvesaq:


jalvesaq:
Support for Typst was added to `knitr` on GitHub on Mar 22, 2026. It's not available in the last released version of `knitr`.

jalvesaq:
My approach in this pull request was to set the filetype of a `.Rtyp` file as `typst` and then enable `R.nvim` features for the Typst filetype. Would it be better to set a new filetype, `Rtyp`, and rename `ftplugin/typst_rnvim.lua` as `ftplugin/rtyp_rnvim.lua`?

lua/r/chunk.lua
```diff
@@ -213,8 +213,69 @@ local get_rnw_code_chunks = function(bufnr)
     return chunks
 end
 
+--- Get code chunks from an RTypst (.Rtyp) document by scanning ```{r} and ```  markers.
```
PMassicotte: Can we not use 

https://github.com/R-nvim/R.nvim/blob/2d0cd152800ced779caaba4684a42efae81203b5/lua/r/chunk.lua#L97-L154

I think rtyps uses the same structure?

I think 

```lua
local params_str = lines[i]:match("^```{r}$")
```

Will not capture params in r chunks, like ```{r echo=FALSE} or ```{python}

PMassicotte:
I think we could just do that in `config.lua`

```lua
vim.treesitter.language.register("markdown", { "quarto", "rmd", "rtyp" })
```

Then we could just reuse `get_rmd_code_chunks` + `parse_info_string_params` + `parse_comment_params all`. My preliminary understanding is that we already have all the tooling to parse `rtyp` code chunks.

What do you think?

lua/r/chunk.lua
jalvesaq: It was easier for me to use regular expressions, but if you could rewrite the function using tree-sitter, it would be great.

PMassicotte:
> My approach in this pull request was to set the filetype of a `.Rtyp` file as `typst` and then enable `R.nvim` features for the Typst filetype. Would it be better to set a new filetype, `Rtyp`, and rename `ftplugin/typst_rnvim.lua` as `ftplugin/rtyp_rnvim.lua`?

Yes, I think this is the right approach.

jalvesaq:
> vim.treesitter.language.register("markdown", { "quarto", "rmd", "rtyp" })

It would not work. Typst is very different from Markdown. It uses `=` instead of `#` for titles, and uses `#` as a prefix for functions in a similar way to LaTeX, which uses `\`.

jalvesaq:
> Yes, I think this is the right approach.

This was my first thought. But then I realized that if the file is a Typst file, we can send the R code to R Console, but we will not be able to process the R code when rendering the PDF because knitr converts a `.Rtyp` file into a `.typ` file. If the file extension is already `.typ`, it would not work because it would overwrite itself.

jalvesaq:
I'll be able to continue working on this only at night...

PMassicotte:
Peek 2026-05-10 08-07

This is what I got so far with TS.

If ok with you, I can push my changes.

PMassicotte:
Note: `r_ls` should be activated for that `filetype`.

PMassicotte:
PS. I have not tested knitting, I do not have the dev version of `knitr` installed.

PMassicotte:
I will push this, let me know if it works for you, otherwise we can just delete the commit if anything :)

PMassicotte:
I am currently using Markdown parser to get the chunks code, inline params and chunks parameters, The Markdown grammar is resilient enough to work with typst filetype. 

I started to implement a typst specific TS extraction code, but we have to do some manually regex job to get the inline parameters in chunks like:

````
```{r, eval=TRUE}
plot(
  x,
  type = "b",
  main = "Random Normal Values",
  xlab = "Index",
  ylab = "Value"
)
```
````

What I have done if more a general refactor on how we get the code chunks based on the file type. At the moment, it works with just using Markdown parser, but maybe I could push the current refactor later in another PR.

**Comment by jalvesaq** [comment]
> If ok with you, I can push my changes.

Yes, it was OK. I think you have already pushed...

**Comment by jalvesaq** [comment]
I'm getting this message when trying to send a line from a chunk of R code:

`Not inside a supported code chunk.`

**Comment by PMassicotte** [comment]
> I'm getting this message when trying to send a line from a chunk of R code:
> 
> `Not inside a supported code chunk.`

Back to the office, I will have a look.

**Comment by jalvesaq** [comment]
I've just found that the filetype `typst` does have the `lang` node if the block is:

````
```r
x <- rnorm(100)
```
````

But not if it is:

````
```{r}
x <- rnorm(100)
```
````

**Comment by PMassicotte** [comment]
I have some meetings this morning, I think I forgot to push a change yesterday. Will do that soon.

**Comment by jalvesaq** [comment]
There is no hurry, and I will only have spare time next weekend.

**Comment by PMassicotte** [comment]
I think this should work now. We might want to have knitr version updated on CRAN before merging in main?

**Comment by jalvesaq** [comment]
Thank you, @PMassicotte and @jiangyun-fun! Everything seems to be Good. Just two notes:

1. We have to run `stylua` to format the files:

```sh
stylua ftplugin/
stylua lua/
```

2. The command `\kr` will not work for Typst, only for RTypst, and the command `\kp` should result in different commands being sent to R depending on the file being `.typ` or `.Rtyp`. Something like this:

```diff
diff --git a/ftplugin/typst_rnvim.lua b/ftplugin/typst_rnvim.lua
index af08276..db59dfe 100644
--- a/ftplugin/typst_rnvim.lua
+++ b/ftplugin/typst_rnvim.lua
@@ -44,7 +53,9 @@ local function build_typst_injections()
     return table.concat(parts, "\n\n")
 end
 
-pcall(vim.treesitter.query.set, "typst", "injections", build_typst_injections())
+if vim.api.nvim_buf_get_name(0):lower():find("%.[Rr][Tt][Yy][Pp]$") then
+    pcall(vim.treesitter.query.set, "typst", "injections", build_typst_injections())
+end
 
 require("r.config").real_setup()
 require("r.rmd").setup()
diff --git a/lua/r/maps.lua b/lua/r/maps.lua
index db6af20..7302531 100644
--- a/lua/r/maps.lua
+++ b/lua/r/maps.lua
@@ -210,7 +210,9 @@ local control = function(file_type)
     create_maps("nvi", "ROBCloseLists",     "r-", "<Cmd>lua require('r.browser').open_close_lists('C')")
 
     if file_type == "typst" then
-        create_maps("nvi", "RMakeRmd",  "kr", "<Cmd>lua require('r.typst').make('default')")
+        if vim.api.nvim_buf_get_name(0):lower():find("%.[Rr][Tt][Yy][Pp]$") then
+            create_maps("nvi", "RMakeRmd",  "kr", "<Cmd>lua require('r.typst').make('default')")
+        end
         create_maps("nvi", "RMakePDFK", "kp", "<Cmd>lua require('r.typst').make('pdf')")
         return
     end
diff --git a/lua/r/typst.lua b/lua/r/typst.lua
index c8becaa..9f1705c 100644
--- a/lua/r/typst.lua
+++ b/lua/r/typst.lua
@@ -4,7 +4,11 @@ M.make = function(outform)
     vim.cmd("update")
     local send = require("r.send").cmd
     if outform == "pdf" then
-        send('knitr::knit2pdf("' .. vim.api.nvim_buf_get_name(0) .. '")')
+        if vim.api.nvim_buf_get_name(0):lower():find("%.[Rr][Tt][Yy][Pp]$") then
+            send('knitr::knit2pdf("' .. vim.api.nvim_buf_get_name(0) .. '")')
+        else
+            send([[system("typst compile ']] .. vim.api.nvim_buf_get_name(0) .. [['")]])
+        end
     else
         send('knitr::knit("' .. vim.api.nvim_buf_get_name(0) .. '")')
     end
```

jalvesaq:
> We might want to have knitr version updated on CRAN before merging in main?

On the one hand, I think it's wiser to wait. On the other hand, we may have to rebase the branch and possibly resolve conflicts if it takes too long for a new version of knitr to be released.

PMassicotte:
Very good comments. 

I am not sure what is going on with `stylua`, why my formatter has not picked it up. I will add a pre-hook to check that.

PMassicotte:
@jalvesaq I added your excellent suggestions and added you as co-author of the commit.

jalvesaq:
Thank you, @PMassicotte!

@gueyenono, could you try the branch from this pull request?

gueyenono:
> Thank you, @PMassicotte!
> 
> @gueyenono, could you try the branch from this pull request?

My apologies. I never saw the notification for this message.
I quickly installed the branch version and everything seems to work fine. It just turns out that R code chunks are all green, so no code highlighting there. The highlighting for Typst works well though.

jalvesaq:
The highlighting of R code is expected to work if the file extension is `.Rtyp` (case insensitive). I've just simplified the pattern used to recognize RTypst, but it was a cosmetic change and shouldn't improve the file type recognition.

What is the extension of your `RTypst` files?

gueyenono:
My file extension is "Rtyp", but I also use Nix. So I'm checking if the issue is not with my configuration.

gueyenono:
Actually I found the issue. I created my code chunk with `{r}` instead of `r`. It all works perfectly now. Thank you.

jalvesaq:
The header of the code chunk should be `{r}` for `.Rtyp` files (which will be converted into `.typ` by `knitr`). The header should be `r` for `.typ` files.

Did you try `<LocalLeader>kp` to generate the PDF from the `.Rtyp` file? It should work with:

````
```{r}
x <- 100
```

The value of `x` is `r x`.
````

but not with:

````
```r
x <- 100
```

The value of `x` is `r x`.
````

**Comment by PMassicotte** [comment]
> Actually I found the issue. I created my code chunk with `{r}` instead of `r`. It all works perfectly now. Thank you.

I am also on NixOS and both work on my side. Not sure what can cause this issue on your side.

**Comment by gueyenono** [comment]
You are right. With `r` as the chunk header, the code in the chunk gets the proper highlighing, but it does execute in the compiled PDF. Also, I did not know that `knitr` also supported actual `typ` files!

**Comment by jalvesaq** [comment]
> You are right. With r as the chunk header, the code in the chunk gets the proper highlighing, but it does execute in the compiled PDF.

I guess you meant "doesn't execute"...

> I did not know that knitr also supported actual typ files!

The development version of `knitr` supports Typst if the extension is `.Rtyp` and the block header is `{r}`, not `r`.

What about this: "It just turns out that R code chunks are all green"? Did you manage to get R code highlighted in RTypst files?

**Comment by gueyenono** [comment]
> > You are right. With r as the chunk header, the code in the chunk gets the proper highlighing, but it does execute in the compiled PDF.
> 
> I guess you meant "doesn't execute"...
> 
> > I did not know that knitr also supported actual typ files!
> 
> The development version of `knitr` supports Typst if the extension is `.Rtyp` and the block header is `{r}`, not `r`.
> 
> What about this: "It just turns out that R code chunks are all green"? Did you manage to get R code highlighted in RTypst files?

1. Yes, I meant it "doesn't" execute
2. Okay, that's what you meant.
3. Actually, my issue was the way I added chunk options:

#```{r plot, out.width="80%"}

#My code is here.

#```
The fact that there was no comma between `r` and `plot` (the chunk label) was the reason why the whole chunk was green. It still ran okay, but there was no highlight. Adding the comma fixes it.

**Comment by jalvesaq** [comment]
> The fact that there was no comma between r and `plot` (the chunk label) was the reason why the whole chunk was green. It still ran okay, but there was no highlight.

I fixed the pattern used to recognize the language of the block, and it now works with a space after the language name, but not with a TAB character.

**Comment by jalvesaq** [comment]
We are listing more tree-sitter parsers in the section _Installation of dependencies_  than are needed by everyone. The parsers for `markdown` and `markdown_inline` are installed by default. So, perhaps we don't need to list them as dependencies. The recommendation could be:

```
   Tree-sitter parsers for `r`, `csv`, and:
         - `yaml` for RMarkdown (`.Rmd`), Quarto (`.qmd`), Rnoweb (`.Rnw`) or
           RTypst (`.Rtyp`) documents.
         - `latex` and `rnoweb` for Rnoweb documents.
         - `typst` for Rtypst documents.
      You can install them using any tree-sitter parser manager (e.g., the
      `nvim-treesitter` or `tree-sitter-manager` plugins) or manually.
      https://github.com/nvim-treesitter/nvim-treesitter
      https://github.com/romus204/tree-sitter-manager.nvim
```

Perhaps even regarding `csv` it could be:

```
         - `csv` for viewing matrices and data frames.
```

jalvesaq:
I documented that not all parsers are always required.

I changed my mind about waiting for the new release of `knitr` to merge this pull request because it doesn't feel good to me to see a pull request that is finished and not merged.

jalvesaq:
I'm going to merge this now...

PR #577, opened 2026-05-16
ci(ci.yml): add stylua job to check Lua code formatting

PMassicotte:
Based on @jalvesaq's comment in #575, I just added a new CI check for code formatting.


PMassicotte:
Fails are expected here, format has been fixed in #575

jalvesaq:
Thank you! I'll merge it.

PR #582, opened 2026-05-21
Enable searching home, project root, and file's directory for latexmkrc file

todmorrison:
This change enables searching home directory, cwd (typically the project root), and then file's directory for a ".latexmkrc" file. It does this in reverse of the stated order, so that the "most local" 'out_dir' configuration stands.

### DIFF
--- a/lua/r/rnw.lua
+++ b/lua/r/rnw.lua
@@ -592,14 +592,20 @@ M.set_pdf_dir = function()
     vim.b.rplugin_pdfdir = "."
 
     -- Latexmk has an option to create the PDF in a directory other than '.'
-    if config.latexcmd and (vim.fn.glob("~/.latexmkrc") ~= "") == 1 then
-        local ltxmk = vim.fn.readfile(vim.fn.expand("~/.latexmkrc"))
-        for _, line in ipairs(ltxmk) do
-            if
-                line:match('out_dir%s*=%s*"(.*)"') or line:match("out_dir%s*=%s*'(.*)'")
-            then
-                vim.b.rplugin_pdfdir = line:match('out_dir%s*=%s*"(.*)"')
-                    or line:match("out_dir%s*=%s*'(.*)'")
+    if config.latexcmd then
+        local chkdirs = { "~", ".", vim.fn.expand("%:p:h") }
+        for _, chkdir in pairs(chkdirs) do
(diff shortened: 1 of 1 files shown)


lua/r/rnw.lua
```diff
@@ -592,14 +592,20 @@ M.set_pdf_dir = function()
...
+                for _, line in ipairs(ltxmk) do
+                    if
+                        line:match('out_dir%s*=%s*"(.*)"')
+                        or line:match("out_dir%s*=%s*'(.*)'")
+                    then
+                        vim.b.rplugin_pdfdir = line:match('out_dir%s*=%s*"(.*)"')
+                            or line:match("out_dir%s*=%s*'(.*)'")
+                    end
+                end
             end
```
PMassicotte:
```suggestion
      for _, chkdir in ipairs(chkdirs) do
          local rcfile = vim.fn.expand(chkdir .. "/.latexmkrc")
          if vim.fn.glob(rcfile) ~= "" then
              for _, line in ipairs(vim.fn.readfile(rcfile)) do
                  vim.b.rplugin_pdfdir = line:match('out_dir%s*=%s*"(.*)"')
                      or line:match("out_dir%s*=%s*'(.*)'")
                      or vim.b.rplugin_pdfdir
              end
          end
      end
```

PMassicotte: Thank you for your contribution.

Two small things:

1. I would use `ipairs`
2. Collapsed the inner `if` into a single assignment (the or `gvim.b.rplugin_pdfdirg` at the end preserves the current value when no match is found)

PMassicotte (changes requested):

PMassicotte:
Note: the CI style errors are expected.

lua/r/rnw.lua
todmorrison: I agree, `ipairs` does exactly what I intended (Thank you!) and the collapsed `if` is much better. I had simply left the existing logic thinking to minimize changes, but your version is so much cleaner.

Changes done. Thanks for the feedback/suggestions.

PMassicotte:
All good on my side :)

jalvesaq:
> All good on my side :)

Thank you, @PMassicotte and @todmorrison! I'm no longer using LaTeX or Rnoweb to try the pull request carefully. Merging it now...

PR #584, opened 2026-05-23
Search for latexmk in more places

todmorrison:
This follows up on PR #582, adding similar logic to search the home directory, cwd (typically the project root), and then file's directory for a ".latexmkrc" file to "nvimcom/R/interlace.R" (the other place where the latexmkrc file i searched for). As in the rmw.lua case, it does this in reverse of the stated order, so that the "most local" 'out_dir' configuration stands.

### DIFF
--- a/nvimcom/R/interlace.R
+++ b/nvimcom/R/interlace.R
@@ -323,7 +323,13 @@ nvim.interlace.rnoweb <- function(
     }
 
     if (!file.exists(logf)) {
-        paths <- c("~/.latexmkrc", ".latexmkrc", paste0(rnwdir, "/.latexmkrc"))
+        paths <- c(
+            paste0(rnwdir, "/.latexmkrc"),
+            ".latexmkrc",
+            "~/.latexmkrc",
+            file.path(Sys.getenv("XDG_CONFIG_HOME"), "latexmk/latexmkrc"),
+            "~/.config/latexmk/latexmkrc"
+        )
         found <- file.exists(paths)
         if (latexcmd == "latexmk" && any(found)) {
             foundlmks <- paths[found]
@@ -336,6 +342,7 @@ nvim.interlace.rnoweb <- function(
                         "/",
                         sub("\\....$", ".log", rnwf)
                     )
+                    break
                 }
             }
         } else {
--- a/nvimcom/R/interlace.R
+++ b/nvimcom/R/interlace.R
@@ -323,15 +323,20 @@ nvim.interlace.rnoweb <- function(
     }
 
     if (!file.exists(logf)) {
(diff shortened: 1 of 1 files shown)


jalvesaq:
It would be good if this pull request could be reviewed by someone who edits Rnoweb files. Section _CONFIGURATION/INITIALIZATION (RC) FILES_ of `latexmk` man page specifies the places where a `latexmkrc` can be found.

jalvesaq:
Looking at the code: shouldn't we have a break at the end of the `if` condition?

```r
if (length(idx) == 1) {
    logf <- paste0(
        sub(".*\\$out_dir\\s*=\\s*['\"](.*)['\"].*", "\\1", lmk[idx]),
        "/",
        sub("\\....$", ".log", rnwf)
    )
    break
}
```

jalvesaq:
The `man` page of `latexmk` lists more places where a `latexmkrc` or `.latexmkrc` can be found than in your pull request...

todmorrison:
> Looking at the code: shouldn't we have a break at the end of the `if` condition?
> 
> ```r
> if (length(idx) == 1) {
>     logf <- paste0(
>         sub(".*\\$out_dir\\s*=\\s*['\"](.*)['\"].*", "\\1", lmk[idx]),
>         "/",
>         sub("\\....$", ".log", rnwf)
>     )
>     break
> }
> ```

No, it shouldn't break in this case. It's scanning the files in reverse order of priority, so the last match should stand.


I do still need to add the other locations mentioned in the man page (and then those additional locations should be added to lua code from the previous PR).

jalvesaq:
> No, it shouldn't break in this case. It's scanning the files in reverse order of priority, so the last match should stand.

It would be more efficient to scan the files in the order of priority and break on the first match.

PMassicotte:
> > No, it shouldn't break in this case. It's scanning the files in reverse order of priority, so the last match should stand.
> 
> It would be more efficient to scan the files in the order of priority and break on the first match.

I think so too :)

todmorrison:
> > > No, it shouldn't break in this case. It's scanning the files in reverse order of priority, so the last match should stand.
> > 
> > 
> > It would be more efficient to scan the files in the order of priority and break on the first match.
> 
> I think so too :)

Fair enough, as long as we're only interested in one variable/setting, searching from the highest priority to the lowest and terminating the search when found is definitely more efficient. (This is the opposite of what latexmk does, however, as it is merging multiple values and the highest priority -- most "local" -- options should override less "local" ones.)

I've reversed the search order and added the break as well as added all the locations latexmk looks for the *user* rc.

jalvesaq:
Thank you! I'll merge it...

PR #587, opened 2026-06-04
New option: `compl_data.max_list_len`

### LINKED ISSUE #586: Building list of objects slow with large unnamed lists
If I create a large unnamed list like so `l <- vector(length = 5e5, mode = "list")` the time to build the list of objects is very long:

Image

The value set in `max_time` is exceeded, shouldn't it be respected by the building process?

Running this example with a minimal init.vim:
```
let mapleader="\<Space>"
let maplocalleader="\<Space>"

call plug#begin(stdpath("data")."/plugged")

Plug 'R-nvim/R.nvim'

Plug 'nvim-treesitter/nvim-treesitter', { 'do': ':TSUpdate', 'branch': 'main'}

call plug#end()

lua <<EOF
local langs = { "r", "markdown", "markdown_inline", "rnoweb", "yaml" } 
vim.api.nvim_create_autocmd("FileType", {
  pattern = langs,
  callback = function()
    vim.treesitter.start()
  end,
})
EOF
```

This is of course a toy example but there is a real use case when loading a large unnamed list which was built from a json object e.g.

Could this be solved by skipping large unnamed lists when building the list of completions?
At least I don't need the completion data for such lists.

This seems to be connected to https://github.com/R-nvim/R.nvim/issues/142.

jalvesaq:
Close #586

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -1338,7 +1338,7 @@ values:
         max_time = 100,
     },
 <
-You should increase the value of `max_list_len` if want to be able to open in
+You should increase the value of `max_list_len` if you want to be able to open in
 the Object Browser and do completion of list and S4 objects with more than
 10000 elements. You should increase the value of `max_depth` if you want to
 complete or see in the Object Browser more than 3 list levels. You should
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -1332,18 +1332,22 @@ in the Lua table `compl_data` in your `R.nvim` config. Below are the default
 values:
 >lua
     compl_data = {
+        max_list_len = 10000,
         max_depth = 3,
         max_size = 1000000,
         max_time = 100,
     },
 <
-You should increase the value of `max_depth` if you want to complete or see in
-the Object Browser more than 3 list levels. You should increase either
-`max_size` or `max_time` if `nvimcom` needs more space or more time to build
-the completion data with the desired list depth. You should decrease the
-values if you notice any delay when R is running commands. In this case,
-please, put `options(nvimcom.verbose = 1)` in your `~/.Rprofile` and use the
-information output by `nvimcom` to decide what parameter to change.
+You should increase the value of `max_list_len` if want to be able to open in
+the Object Browser and do completion of list and S4 objects with more than
+10000 elements. You should increase the value of `max_depth` if you want to
+complete or see in the Object Browser more than 3 list levels. You should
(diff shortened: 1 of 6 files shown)


nvimcom/src/nvimcom.c
```diff
@@ -1272,11 +1276,12 @@ static void *client_loop_thread(__attribute__((unused)) void *arg)
  *
  * @param rinfo Information on R to be passed to nvim.
  */
-SEXP nvimcom_Start(SEXP vrb, SEXP anm, SEXP swd, SEXP imd, SEXP szl, SEXP tml,
-                   SEXP dbg, SEXP nvv, SEXP rinfo) {
+SEXP nvimcom_Start(SEXP vrb, SEXP anm, SEXP swd, SEXP imd, SEXP lsl, SEXP szl,
```
PMassicotte: I think there is an error in the order of the parameters

Should it be:

```c
SEXP nvimcom_Start(SEXP vrb, SEXP anm, SEXP swd, SEXP lsl, SEXP imd, SEXP szl, ...)
```

instead of 

```c
SEXP nvimcom_Start(SEXP vrb, SEXP anm, SEXP swd, SEXP imd, SEXP lsl, SEXP szl, ...)
```

doc/R.nvim.txt
```diff
@@ -1332,18 +1332,22 @@ in the Lua table `compl_data` in your `R.nvim` config. Below are the default
...
     },
 <
-You should increase the value of `max_depth` if you want to complete or see in
-the Object Browser more than 3 list levels. You should increase either
-`max_size` or `max_time` if `nvimcom` needs more space or more time to build
-the completion data with the desired list depth. You should decrease the
-values if you notice any delay when R is running commands. In this case,
-please, put `options(nvimcom.verbose = 1)` in your `~/.Rprofile` and use the
-information output by `nvimcom` to decide what parameter to change.
+You should increase the value of `max_list_len` if want to be able to open in
```
PMassicotte:
```suggestion
You should increase the value of `max_list_len` if you want to be able to open in
```

PMassicotte:
Two small comments, other than that it is perfect, good job!

nvimcom/src/nvimcom.c
jalvesaq: You are right!

jalvesaq:
Thank you for the review, @PMassicotte! I think everything is OK now, and I will merge it.

PR #589, opened 2026-06-05
Only warn of parsers needed by the current filetype

jalvesaq:


jalvesaq:
The documentation and `:checkhealth r` only require the `r` parser, but we check for most parsers on startup. With this pull request, we only check for the parsers required by the current file type.

jalvesaq:
`:RDebugInfo` was failing to correctly detect the language in the YAML header of a Quarto document as `yaml` and normal text as `markdown_inline`. I deleted these two checks.

lua/r/config.lua
```diff
@@ -1360,22 +1359,35 @@ M.check_health = function()
...
+    elseif vim.bo.filetype == "rnoweb" then
+        table.insert(needed, "rnoweb")
+    elseif vim.bo.filetype == "typst" then
+        table.insert(needed, "typst")
+    end
+    for _, v in pairs(needed) do
+        if not has_parser(v, parsers) then
+            swarn(
+                string.format(
+                    'R.nvim requires the following tree-sitter parsers: "%s". Please, install them.',
```
PMassicotte: Just one small thing

When `v` is the missing parser, `table.concat(needed, ...)` still concatenates the entire needed list, not just the missing one. ie. all parser in the table will be printed. I think?

PMassicotte (commented):
All good! Just this message, I think it will print all the needed parsers, not just the missing one.

lua/r/config.lua
jalvesaq: Initially, the `needed` is only `{"r"}`. The other parsers are inserted into the table according to the current value of `filetype`. So, they are all needed for the specific file being edited.

jalvesaq:
> All good! Just this message, I think it will print all the needed parsers, not just the missing one.

I see your point now. It might be better to be specific and warn about the missing one that triggered the warning.

PMassicotte:
All good with me, maybe ok also to print all. I do not have preference in fact. Lets merge it?

PMassicotte:
Thank you :+1:

PR #594, opened 2026-06-16
New nvimcom option: `max_cpu_cores`

### LINKED ISSUE #593: Update detection of CPU count
Thanks for R.nvim!

Currently, this uses the `parallel::detectCores()`, to get the number of cores. This gets *all CPUs on the hardware*, which causes awkward situations when running on an HPC cluster. For example, if I request 2 CPUs on an interactive node, R.nvim grabs all 128 cores on the node, and the fighting it causes with the scheduler causes the node to hang (and get an admonishing email from the HPC staff). It also uses a surprising amount of memory (19.8 GB by the time I killed it).

Using the `future::availableCores()` should prevent this, e.g., change [this line](https://github.com/R-nvim/R.nvim/blob/2d42670c3f73098411db4f09961fcda288ca271e/nvimcom/R/bol.R#L693):

```diff
-num_cores <- max(c(parallel::detectCores() - 2, 1))
+num_cores <- max(c(future:::availableCores() - 2, 1)) 
```

It does require an additional dependency, but with no way to disable completion data collection like that proposed in #522 (in [this comment](https://github.com/R-nvim/R.nvim/issues/522#issuecomment-4104391366)) and grabbing all CPUs on a node, I don't think it's possible to use R.nvim on a cluster.

jalvesaq:
Close #593

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -1351,20 +1351,18 @@ change.
 
 Completion lists are built in parallel, using all but two of the cores
 reported by `parallel::detectCores()`. On systems where this number does not
-reflect the resources actually available to the process (for example, an HPC
-node managed by a job scheduler such as Slurm), you can cap the number of
-cores with the `nvimcom.max_cpu_cores` option in your ~/.Rprofile:
+reflect the resources actually available to the process, like an HPC
+cluster where CPUs are allocated as portions of the total available on the hardware,
+you can cap the number of cores with the `nvimcom.max_cpu_cores` option in your ~/.Rprofile:
 >r
    options(nvimcom.max_cpu_cores = 4)
 
-In case you are using Slurm, you can set the option to the value of the
-environment variable `SLURM_CPUS_PER_TASK`:
->r
-   options(nvimcom.max_cpu_cores = as.numeric(Sys.getenv("SLURM_CPUS_PER_TASK")))
+This value can come from environment variables. For example, on an HPC cluster using
+the Slurm scheduler, you can set the option to the value of the environment variable `SLURM_CPUS_PER_TASK` or, if it's not available (for example, you have not specified
+`cpus-per-task` for the job and only get the default number of CPUs), then fall back to 2:
 
-Note that `SLURM_CPUS_PER_TASK` is only set if you have requested more than
-one CPU core in your Slurm job script (e.g., with `#SBATCH
---cpus-per-task=4`).
+>r
+   options(nvimcom.max_cpu_cores = max(2, as.numeric(Sys.getenv("SLURM_CPUS_PER_TASK"))))
 
 It is possible to add a custom keymap through `objbr_mappings` table. The
 keymap defined in this table will be available in the object browser only.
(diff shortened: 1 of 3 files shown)


PMassicotte:
Good job. Should we also give an example about using `SLURM` requested CPUs?

```r
options(nvimcom.max_cpu_cores = as.integer(Sys.getenv("SLURM_CPUS_PER_TASK"))
```

jalvesaq:
> Should we also give an example about using SLURM requested CPUs?

Yes, if the variable will always be set for Slurm users in the normal way of using Slurm (which, according to Gemini, isn't through SSH). I have zero experience with HPC Clusters.

PMassicotte:
I will add something.

PMassicotte:
I think this is good to merge. Ok with you?

jalvesaq:
It's OK for me. Should we wait a while for @daler's comments?

PMassicotte:
> It's OK for me. Should we wait a while for @daler's comments?

I think so!

doc/R.nvim.txt
```diff
@@ -1349,6 +1349,23 @@ this case, please, put `options(nvimcom.verbose = 1)` in your `~/.Rprofile`
...
+
+In case you are using Slurm, you can set the option to the value of the
+environment variable `SLURM_CPUS_PER_TASK`:
+>r
+   options(nvimcom.max_cpu_cores = as.numeric(Sys.getenv("SLURM_CPUS_PER_TASK")))
+
+Note that `SLURM_CPUS_PER_TASK` is only set if you have requested more than
+one CPU core in your Slurm job script (e.g., with `#SBATCH
+--cpus-per-task=4`).
+
```
daler, line 1366:
```suggestion
Completion lists are built in parallel, using all but two of the cores
reported by `parallel::detectCores()`. On systems where this number does not
reflect the resources actually available to the process, like an HPC
cluster where CPUs are allocated as portions of the total available on the hardware,
you can cap the number of cores with the `nvimcom.max_cpu_cores` option in your ~/.Rprofile:
>r
   options(nvimcom.max_cpu_cores = 4)

This value can come from environment variables. For example, on an HPC cluster using
the Slurm scheduler, you can set the option to the value of the environment variable `SLURM_CPUS_PER_TASK` or, if it's not available (for example, you have not specified
`cpus-per-task` for the job and only get the default number of CPUs), then fall back to 2:

>r
   options(nvimcom.max_cpu_cores = max(2, as.numeric(Sys.getenv("SLURM_CPUS_PER_TASK"))))

```

daler (commented):
Tested, and works great -- thanks for such quick turnaround.

I made some adjustments to the docs to make it a little more general while still keeping the concrete Slurm-specific example, and slightly tweaked the example to have a max of 2 in case the env var is unavailable, like if you don't use `--cpus-per-task`.

jalvesaq:
Thank you, @daler! Let's wait for @PMassicotte final review.

PMassicotte:
LGTM! Merging, thank you all for your contributions.

PR #596, opened 2026-06-23
refactor(references.lua): move pipe chain column resolution logic before target definition check

PMassicotte:
Small bug fix when using *renaming symbol* in a pipe expression.

```r
summarise_mtcars <- function(data = mtcars) {
  data |>
    as_tibble(rownames = "model") |>
    mutate(power_to_weight = hp / wt) |>
    summarise(
      mean_mpg = mean(mpg),
      mean_p2w = mean(power_to_weight),
      .by = cyl
    )
}
```

Place the cursor on `power_to_weight` in the `mutate()` function. Try to rename, and only the `power_to_weight` in the `summarise()` gets renamed. This PR make sure that both are renamed.

### DIFF
--- a/lua/r/lsp/references.lua
+++ b/lua/r/lsp/references.lua
@@ -76,19 +76,18 @@ function M.find_locations(line, col, bufnr)
         return { locations = locs, word = word, resolved = false }
     end
 
+    local pipe = require("r.lsp.pipe")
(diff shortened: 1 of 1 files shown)


jalvesaq:
Thank you! I confirm the bug and that the pull request fixes it!

PR #597, opened 2026-06-23
Fix text width of documentation

jalvesaq:


jalvesaq:
Reflow some paragraphs with lines larger than 78 characters in the documentation.

PMassicotte:
Tx!

PR #599, opened 2026-06-27
fix(config): handle Windows rnvim config path

hongyuanjia:
## Summary

Fix the per-buffer `rnvim_config.lua` lookup so it works with Windows buffer paths. The previous slash-only Lua pattern could return `nil` for backslash-separated paths and fail before checking whether the config file exists.

This bug was introduced in PR #515.

## Changes

- Build the config path with `vim.fs.dirname()` and `vim.fs.joinpath()` instead of a hard-coded `/` pattern.
- Guard the lookup when the current buffer has no file name.

### DIFF
--- a/lua/r/config.lua
+++ b/lua/r/config.lua
@@ -1289,10 +1289,13 @@ M.real_setup = function()
 
     require("r.rproj").apply_settings(config)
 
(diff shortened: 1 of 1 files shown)


jalvesaq:
Thank you!

PR #600, opened 2026-06-27
New option: `rconsole_pos`

### LINKED ISSUE #598: Console split ignores `nosplitright` due to hardcoded `belowright` in builtin.lua
The documentation suggests setting `nosplitright` if you want the R Console on the left side. However, this has no effect, the console always opens on the right.
 
The cause is in `lua/r/term/builtin.lua`, where `split_window()` hardcodes `belowright` in its split commands:
 
```lua
vim.cmd("silent exe 'belowright " .. config.rconsole_width .. "vnew'")
```
 
Since `belowright` (like `aboveleft`/`topleft`) explicitly overrules `splitright`/`splitbelow`, the user's `nosplitright` setting is never consulted. The same applies to the other `belowright` splits in that function (the `vnew` fallback and the horizontal `new` cases).
 
It would be nice to either respect `splitright`/`splitbelow`, or expose a config option to control console placement.

jalvesaq:
Close #598

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -976,7 +976,6 @@ You may close the window opened by `:RConfigShow` by simply pressing `q`.
 |rconsole_height|     Number of lines of R Console
 |rconsole_width|      Number of columns of R Console
 |rconsole_pos|        Position of the R Console relative to the editor window
-|register_treesitter| Register treesitter parsers for quarto and rmd files
 |min_editor_width|    Minimum number of columns of editor after R start
 |RStudio_cmd|         Run RStudio instead of R.
 |listmethods|         Do `nvim.list.args()` instead of `args()`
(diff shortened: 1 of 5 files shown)


jalvesaq:
Although we are in a freeze period, this new option was necessary to fix a bug in the documentation and restore a feature of Vim-R (previously known as Nvim-R).

The `rconsole_pos` works for the built-in terminal and for WezTerm.

For Tmux split, the user has to set the `tmux` command. Hence, it's up to the user to add the `-b` parameter to get a split pane on the left or above the current one. For Kitty split, I didn't find a parameter to control the position of the new window.

doc/R.nvim.txt
```diff
@@ -975,6 +975,7 @@ You may close the window opened by `:RConfigShow` by simply pressing `q`.
 |config_tmux|         Don't use a specially built Tmux config file
 |rconsole_height|     Number of lines of R Console
 |rconsole_width|      Number of columns of R Console
+|rconsole_pos|        Position of the R Console relative to the editor window
```
PMassicotte, line 978: Should this also be added in 6.14? Along with

 *rconsole_height*
*rconsole_width*
 *rconsole_pos*
   *min_editor_width*

PMassicotte:
All good on my side; tested and worked perfectly; thank you! Maybe just add the missing `rconsole_pos` anchor in 6.14?

doc/R.nvim.txt
jalvesaq: I'll add it...

PMassicotte:
Thank you very much!

PR #602, opened 2026-06-29
refactor: use builtin path functions

kkanden:
this PR aims to unify path manipulation around the codebase both to ensure consistency and compatibility on all platforms using `vim.fs` and `vim.uv` function where appropriate. 

i'll make commits limited to files or in larger batches to make debugging and reviewing easier.

### DIFF
--- a/lua/r/edit.lua
+++ b/lua/r/edit.lua
@@ -134,7 +134,7 @@ M.vim_leave = function()
     end
 
     for _, fn in pairs(del_list) do
-        vim.fn.delete(fn)
+        vim.fs.rm(fn, { force = true })
     end
 
     if config.remote_R_host == "" then
@@ -501,7 +501,7 @@ M.open_example = function()
     vim.api.nvim_set_option_value("bufhidden", "wipe", { scope = "local" })
     vim.api.nvim_set_option_value("swapfile", false, { scope = "local" })
     vim.api.nvim_set_option_value("buftype", "nofile", { scope = "local" })
-    vim.fn.delete(config.tmpdir .. "/example.R")
+    vim.fs.rm(vim.fs.joinpath(config.tmpdir, "example.R"), { force = true })
 end
 
 return M
--- a/lua/r/edit.lua
+++ b/lua/r/edit.lua
@@ -448,7 +448,7 @@ M.view_df = function(oname, txt)
     if config.view_df.save_fun and config.view_df.save_fun ~= "" then
         tsvnm = txt
     else
-        tsvnm = config.tmpdir .. "/" .. oname .. ".tsv"
+        tsvnm = vim.fs.joinpath(config.tmpdir, oname .. ".tsv")
         vim.fn.writefile(csv_lines, tsvnm)
     end
     M.add_for_deletion(tsvnm)
@@ -480,21 +480,28 @@ end
 M.open_example = function()
     local bl = vim.api.nvim_list_bufs()
     for _, v in pairs(bl) do
-        if vim.api.nvim_buf_get_name(v) == config.tmpdir .. "/example.R" then
+        if
+            vim.fs.normalize(vim.api.nvim_buf_get_name(v))
+            == vim.fs.joinpath(config.tmpdir, "example.R")
+        then
             vim.cmd("bunload! " .. tostring(v))
             break
         end
     end
 
     if config.nvimpager == "tabnew" or config.nvimpager == "tab" then
-        vim.cmd("tabnew " .. config.tmpdir:gsub(" ", "\\ ") .. "/example.R")
+        vim.cmd("tabnew " .. vim.fs.joinpath(config.tmpdir:gsub(" ", "\\ "), "example.R"))
     else
         if config.nvimpager == "split_v" then
             vim.cmd(
-                "belowright vsplit " .. config.tmpdir:gsub(" ", "\\ ") .. "/example.R"
+                "belowright vsplit "
+                    .. vim.fs.joinpath(config.tmpdir:gsub(" ", "\\ "), "example.R")
             )
         else
-            vim.cmd("belowright split " .. config.tmpdir:gsub(" ", "\\ ") .. "/example.R")
+            vim.cmd(
+                "belowright split "
+                    .. vim.fs.joinpath(config.tmpdir:gsub(" ", "\\ "), "example.R")
+            )
         end
     end
     vim.api.nvim_buf_set_keymap(0, "n", "q", ":q<CR>", { noremap = true, silent = true })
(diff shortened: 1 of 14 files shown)


kkanden:
i'll try to get the rest done tomorrow or later if there's not enough time. many of the path joins using `..` are probably fine once we have ensured that paths are normalized but for the sake of consistency i'm gonna try to turn everything into `vim.fs.joinpath`. if you'd like for me to leave the simple path joins as is then let me know :)

kkanden:
i think this covers it all. please do let me know if i wasn't too eager with the use of `vim.fs.joinpath`.

i wasn't too sure what to do in rnw.lua#355 and rnw.lua#396 (`SyncTex_get_master` and `SyncTex_backward`). it seems to me there's some synctex-specific manipulation being done so i left it as is. after that is resolved the `normalize_windows_path` function from utils could be retired in favor of just `vim.fs.normalize`.

PMassicotte:

A few points to validate:

- `vim.fs.rm` throws an error if the file is already gone, i.e., quitting R and then Neovim causes the exit cleanup loop to crash mid-way, leaving temp files behind.
- If R crashes before it finishes starting up, `clear_R_info` tries to delete files that were never created, throws, and skips resetting R.nvim's internal state.
- A user with `compldir = "$HOME/.local/R.nvim"` in their config gets a literal directory named `$HOME` created in their cwd instead of their actual home directory path.

PMassicotte:
To verify:  `M.normalize_path(path) return vim.fs.normalize(path) end`

`vim.fs.normalize tidies up a path` but it does not make relative paths absolute. Pass `foo.R` and you get back `foo.R`.

The function is used in `workspace.lua` and `references.lua` to build cache keys and compare file locations. Have to make sure it does not break things with the LSP.

kkanden:
thanks a lot for the feedback!
- ad. `vim.fs.rm`: we can either stick to `vim.fn.delete` for the sake of simplicity or we could make a wrapper that first checks if the file exists and then delete it. i'm fine with either.
- ad. `compldir`: i'm pretty sure that is caused by executing `set_directories` (inside `do_common_global`) *before* user options are merged with the default config (`config.lua` lines ~ 1198-1208). from what i've tested, environment variables are not expanded on the main branch either.
- ad. `normalize_path`: that is an oversight on my side. i can wrap `fnamemodify` in `fs.normalize` or leave it as is since it seems to work correctly.

jalvesaq:
I will be able to try the changes only on Saturday. For now, just two observations after quickly looking at the changes:

`sumatra.lua` is used only on Windows because SumatraPDF is a Windows-only application. Hence, it's simpler to hardcode the backslash.

In code such as this:

```lua
vim.fs.joinpath(config.remote_compl_dir, "tmp/bo_code.R")
```

there is a forward slash that isn't being normalized.

kkanden:
as per `:help vim.fs.joinpath` slashes in the arguments are normalized:
```
vim.fs.joinpath({...})                                     *vim.fs.joinpath()*
   [...]
    Examples:
    • "foo/", "/bar" => "foo/bar"
    • "", "after/plugin" => "after/plugin"
    • Windows: "a\foo\", "\bar" => "a/foo/bar"
```

kkanden:
i have reverted the change in lsp/utils.lua and some of the path joins in pdf/sumatra.lua. for the rest of the issues, i'll wait for your opinion on how to resolve them best :)

PMassicotte:
Hi and thank you. I will be back to my office tomorrow; I will have more time to dig in :)

lua/r/edit.lua
```diff
@@ -134,7 +134,7 @@ M.vim_leave = function()
     end
 
     for _, fn in pairs(del_list) do
-        vim.fn.delete(fn)
+        vim.fs.rm(fn)
```
PMassicotte:
```suggestion
        vim.fs.rm(fn, { force = true })
```
I think this should be changed everywhere.

PMassicotte (changes requested):

PMassicotte:
It seems to be working on my side, but I do not have access to a Windows machine. @jalvesaq, have you ?

jalvesaq:
>  It seems to be working on my side, but I do not have access to a Windows machine. @jalvesaq, have you ?

No. I uninstalled the virtual machine months ago because it was too slow.

PMassicotte:
> > It seems to be working on my side, but I do not have access to a Windows machine. @jalvesaq, have you ?
> 
> No. I uninstalled the virtual machine months ago because it was too slow.

Maybe we could use CI to do at least some basic checks on a Windows machine.

kkanden:
i used this branch a but on my windows machine and encountered no issues. couldn't test the whole functionality tho.

PMassicotte:
@kkanden thank you for the changes. Are you working on Windows? If so, do you have any issue? I am asking because `start_rnvimserver` uses ";" when `config.is_windows` is true, but MinGW/MSYS2 shells use ":" as the PATH separator.

kkanden:
my work laptop is on windows and in `vim.uv.os_uname()` it shows up as mingw32. the lsp worked fine despite that.

jalvesaq:
I'm not using R these days, but I tried the pull request quickly and couldn't find any bugs while using `R.nvim`. Looking at the code, I only noted a change that seems strange. An opening single quote was replaced with a closing single quote:

```diff
--- a/lua/r/rdoc.lua
+++ b/lua/r/rdoc.lua
@@ -67,9 +67,9 @@ M.fix_rdoc = function(txt)
     txt = string.gsub(txt, "\018", "\\")
     txt = string.gsub(txt, "<URL: %([^>]*%)>", " |%1|")
     txt = string.gsub(txt, "<email: %([^>]*%)>", " |%1|")
-    if not config.is_windows then
+    if not config.is_windows or config.is_mingw then
         -- curly single quotes only if the environment is UTF-8
-        txt = string.gsub(txt, "\145", "‘")
+        txt = string.gsub(txt, "\145", "’")
         txt = string.gsub(txt, "\146", "’")
     end
```

PMassicotte:
wow good catch! I had to look at your diff for 1 min to get the issue :)

jalvesaq:
I'll ask Copilot to make a final review, and then we can merge it.

PMassicotte:
A lot of good suggestions here. Like avoiding manual `/`.

jalvesaq:
Yes, but see https://github.com/R-nvim/R.nvim/pull/602#issuecomment-4848291898

jalvesaq:
Regarding SyncTeX, I would only touch the code if there were a bug report. And if the bug is specific to Windows, someone would have to make a pull request to fix it.

jalvesaq:
I had forgotten to try SyncTeX yesterday, but I tried it now, and it's working on Linux.

jalvesaq:
Is this pull request ready to be merged?

kkanden:
it's ready from my side :)

PMassicotte:
I think so!

jalvesaq:
Thank you, @kkanden and @PMassicotte !

kkanden:
thank you guys for the amazing plugin :D

PR #604, opened 2026-07-02
fix(bol.R): set num_cores to 1 on Windows to prevent parallel execution issues (closes #603)

### LINKED ISSUE #603: On windows, the R console doesn't load - seems to be assuming forking possible?
Hi there,

I think compat with W11 is broken at the moment due to the implicit assumption that forking processes is possible via use of `parallel::mclapply` at some point in the setup process.

`nvimcom` builds correctly without any errors, but this could potentially still be an issue with the `R` package I suppose, and therefore an issue for them. If so, apologies. If not, here are some diagnostics to help. I've tried to set `num_cores` in config but not sure where to place it. 

## Issue

When I `<localLeader>rf` to start the console, I get `notify.warn R.nvim "R" exited with status 1`, which isn't very helpful in diagnosing the issue. Then, exactly 1 minute later I get `notify.warn R.nvim The package nvimcom wasn't loaded yet. Please, quit R and try again.`

I can still function by circumventing `nvimcom` completely by `<leader>ft` to open a terminal panel, start `R` there and manually copy paste lines of code to run, but it's a lot less convenient and I've found myself having to go back to Positron/Rstudio to run things on windows that I can't run in WSL (e.g. python bridge to Excel via `xlwings`, which is far better than `openxlsx2` and allows direct control, but is windows-only and doesn't work at all via `linux`). 

From the below it seems to be something to do with a change to the codebase around parallel processing. Perhaps a compat layer was taken out at some point which used psockets in favour of mclapply, or maybe something that was serial only was replaced with it?

Either way, windows implementations are way slower anyway, so hopefully an easy fix would be a platform check and just use `lapply` on W11? As long as it works, I don't mind waiting a couple more seconds!

Thanks, and hopefully it's an easy fix!

## Config and diagnostic outputs

current config is the default for `LazyVim`, with this lua block to direct to the right R version, as there are multiple on my machine (standard windows issue...):
```lua
return {
  {
    "R-nvim/R.nvim",
    opts = {
      R_path = "C:/Program Files/R/R-4.6.1/bin",
    },
  },
}
```

`RDebugInfo`:
```
08:11:38 msg_show.echo Time:
  global setup: 0.0085424
  pdf setup (async): 1.39e-05
  windows setup: 0.0001642
  rmd setup: 0.0012813
  before_rns.R (async): 1.6568903
  make rnvimserver: 0.3685229
  check health (async): 0.0089565
before_rns.R stdout: (empty)
before_rns.R stderr: (empty)
stderr of last completion data building: Error in parallel::mclapply(1:nrow(b), process_row, mc.cores = num_cores) :   'mc.cores' > 1 is not supported on WindowsCalls: <Anonymous> -> <Anonymous>
Execution halted

Language detection in a Quarto document:
  Correctly detected: markdown
  Correctly detected: r
Language detection in an Rnoweb document:
  Correctly detected: latex
  Correctly detected: chunk_header
  Correctly detected: r
  Correctly detected: chunk_end
```

`sessionInfo()` after opening a console with `<leader>ft` and typing `R`:
```
R version 4.6.1 (2026-06-24 ucrt)
Platform: x86_64-w64-mingw32/x64
Running under: Windows 11 x64 (build 26200)

Matrix products: default
  LAPACK version 3.12.1

locale:
[1] LC_COLLATE=English_United Kingdom.utf8 
[2] LC_CTYPE=English_United Kingdom.utf8
[3] LC_MONETARY=English_United Kingdom.utf8
[4] LC_NUMERIC=C
[5] LC_TIME=English_United Kingdom.utf8

time zone: Europe/London
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

other attached packages:
[1] nvimcom_0.9.96

loaded via a namespace (and not attached):
[1] compiler_4.6.1 parallel_4.6.1 tools_4.6.1
```

PMassicotte:
`parallel::mclapply` does not seem to be working on Windows, so we force the number of cores to be 1.

### DIFF
--- a/nvimcom/R/bol.R
+++ b/nvimcom/R/bol.R
@@ -695,6 +695,11 @@ nvim.build.cmplls <- function() {
     } else {
         num_cores <- getOption("nvimcom.max_cpu_cores")
     }
+
+    if (.Platform$OS.type == "windows") {
+        num_cores <- 1
+    }
+
     invisible(parallel::mclapply(1:nrow(b), process_row, mc.cores = num_cores))
 
     return(invisible(0))


PMassicotte:
I am surprised that this issue (#604) was not reported before. @jalvesaq have you experienced this?

jalvesaq:
I don't remember if I still had the virtual machine when we started using parallel execution. Maybe everyone is using WSL. For many, it might be simpler to stick with Vim and Vim-R, rather than migrate to Neovim and R.nvim. This would explain the existence of bugs and the absence of bug reports.

In my opinion, you should not dedicate too much of your time trying to fix bugs that don't affect you. I did this for years, and the result is that I'm tired and want to mostly retire from the development of R.nvim.

PMassicotte:
> In my opinion, you should not dedicate too much of your time trying to fix bugs that don't affect you. I did this for years, and the result is that I'm tired and want to mostly retire from the development of R.nvim.

Ah I am sorry to ear that. I know how it can feels. But we need you!

richardsc:
> In my opinion, you should not dedicate too much of your time trying to fix bugs that don't affect you. I did this for years, and the result is that I'm tired and want to mostly retire from the development of R.nvim.

> Ah I am sorry to ear that. I know how it can feels. But we need you!

Just to second @PMassicotte here -- R.nvim has been a huge part of my coding workflow for the last two years, after switching from >15 years using Emacs+ESS. As a maintainer of an R package that is extensive, fairly popular, and has a diverse range of users (the [`oce package`](https://github.com/dankelley/oce)) I can understand the fatigue of chasing other people's bugs and requests.

It's been a goal of mine to better learn the R.nvim internals (and lua) to contribute back when possible, and this is motivation for me to try and find some time to dive in more.

PR #608, opened 2026-07-07
feat: implement `nvimpager = "float"`

kkanden:
this implements the "float" option for nvimpager for both R help and example. i have added a utils function for creating floating.

in the future, a user config option could be added to customize the location and size of the floating window. by default it just centers the window.

### DIFF
--- a/doc/R.nvim.txt
+++ b/doc/R.nvim.txt
@@ -1514,7 +1514,8 @@ The valid values of `nvimpager` are:
                  very top of the screen.
    "split_v"   : Split the window vertically. Display R documentation to
                  the right of the editor window.
-   "float"     : Not implemented yet.
+   "float"     : Show the help documentation in a float window centered on the
+                 editor window.
    "no"        : Do not show R documentation in Neovim.
 
 
--- a/lua/r/doc.lua
+++ b/lua/r/doc.lua
@@ -5,6 +5,7 @@ local utils = require("r.utils")
 local cursor = require("r.cursor")
 local job = require("r.job")
 local doc_buf_id = nil
+local doc_win_id = -1 -- for `nvimpager = "float"`
 
 local M = {}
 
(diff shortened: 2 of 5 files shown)


jalvesaq:
Thank you! It works!

Some ideas (waiting for comments):

- Limit the maximum width to a not too high value, such as 120?

- Prefer 80 columns as the width of the float window?

- Between 60 and 120 columns, choose a value corresponding to the largest line in the document +1?

- Position of the float window centered on the editor window, not on the full Neovim display?

kkanden:
good ideas, i'll try to implement these changes later today and see what looks good

kkanden:
i have implemented the changes:
- floating window is centered relative to the code editor
- if 60 <= cols <= 120 then the width is longest line length + 1, 80 otherwise

one shortcoming (or feature?) compared to the last version is that the example/help window does not get directly overlaid on top of the preview help/example window, but is relative to it. subsequent help->help/example->example windows are overlaid as before.

jalvesaq:
Thank you, @kkanden!

Some observations:

- Some variables in `utils.open_float_win()` are always overridden (`width`, `height`, `col`, and `row`), and the same code to calculate them is repeated in `edit.lua` and `doc.lua`. Hence, you can:
  - Pass `lines` to `utils.open_float_win()` and let the values of `width`, `height`, `col`, and `row` be calculated there.
  - Delete the code that calculates the variables from `edit.lua` and `doc.lua`.
- If the Object Browser is open, the float window will not be centralized in the editor window.

kkanden:
>     * Some variables in `utils.open_float_win()` are always overridden (`width`, `height`, `col`, and `row`), and the same code to calculate them is repeated in `edit.lua` and `doc.lua`. Hence, you can:
>       
>       * Pass `lines` to `utils.open_float_win()` and let the values of `width`, `height`, `col`, and `row` be calculated there.
>       * Delete the code that calculates the variables from `edit.lua` and `doc.lua`.
so in the `open_float_win` function we want the window to always be calculated from the line lengths of the text to be displayed (given the previous conditions)? i thought of the function to be a generic utils function.
>     * If the Object Browser is open, the float window will not be centralized in the editor window.
that happens if the editor window is less than 60 and the width of the floating window defaults to 80 columns. this also makes the window look like it's stuck to the side of the editor instead of centered. i thought this looked better than a centered but small and squished window.

jalvesaq:
I'm sorry, but I'm on a trip now and will need some time to properly review the code again...

jalvesaq:
The pull request is ready for review again (although the "format-suggest/yaml" check is failing).

jalvesaq:
I totally messed up the git history of the pull request when rebasing the branch on the main branch. The only way that I know to fix it is to create a new branch and make a new pull request, but this would erase @kkanden from the git history. What should we do? Merge the branch as is (and fix the merge description while merging)? Or could @kkanden make a new pull request from a new branch, overwriting some files with the files from this branch?

kkanden:
> I totally messed up the git history of the pull request when rebasing the branch on the main branch. The only way that I know to fix it is to create a new branch and make a new pull request, but this would erase @kkanden from the git history. What should we do? Merge the branch as is (and fix the merge description while merging)? Or could @kkanden make a new pull request from a new branch, overwriting some files with the files from this branch?

if it's okay with you, i don't have a problem with not appearing in the git history. if not, what would need to be done on the new branch to get the pr in good state?

jalvesaq:
Thankyou! I'll try to merge this pull request. I think it will work. We can fix any bugs later...

PR #610, opened 2026-07-10
fix(Makefile): ensure CC is set to gcc only if not already defined (closes #606)

### LINKED ISSUE #606: rnvimserver: build fails because GNU make defaults CC to cc
Building rnvimserver with GNU can fail if CC is not set in the environment or if there is no "cc" the path that maps to a gcc-compatible compiler.

The following line in the Makefile fails to redefine CC as it is defined by default in GNU make to be "cc" (see [info make](https://manpagez.com/info/make/make-4.4.1/Implicit_Variables.php)). 

https://github.com/R-nvim/R.nvim/blob/c56ebe0f8445e251673981c40ac2d74659ecd6ed/rnvimserver/Makefile#L1

If CC is not defined by user, $(CC) will always evaluate to ```cc```, and without a "cc" in the path, the build for rnvimserver will fail with the error ```cc: No such file or directory```.

Just to confirm, I ran ```make``` on this simple test Makefile

```make
test:
  echo $(CC)
```

and on both linux and windows it prints

```
echo cc
cc
```

It is possible that this could be a considered a windows-specific error, as a "cc" seems to always be included with gcc on linux. In my particular case on windows, I am using the build tools from rtools45, and it does not provide a companion cc for its gcc.

Is the purpose of the ```CC =? gcc``` check to allow users to set their preferred compiler in their environmental variables? It might need to be replaced with something more robust.

EDIT: The simplest fix for me was to just define CC in my environmental variables.

PMassicotte:
See #606.

### DIFF
--- a/rnvimserver/Makefile
+++ b/rnvimserver/Makefile
@@ -1,4 +1,6 @@
-CC ?= gcc
+ifeq ($(origin CC),default)
+CC = gcc
+endif
 SRCS = complete.c resolve.c hover.c definition.c signature.c rhelp.c chunk.c roxygen.c data_structures.c logging.c rnvimserver.c obbr.c tcp.c utilities.c ../nvimcom/src/common.c
 
 ifeq ($(OS),Windows_NT)


jalvesaq:
I can't review the pull request because I don't have access to Windows.

Note: in `doc/R.nvim.txt`, we say that `R.nvim` requires "a C compiler (e.g. `gcc` or `clang`)".

jalvesaq:
@ctm22396, can you comment on this pull request?

ctm22396:
Yes, this resolves the issue. For reference, here are the results of the test I ran on the branch

https://github.com/R-nvim/R.nvim/issues/606#issuecomment-4936840167

PR #615, opened 2026-08-19
fix(flake): compile rnvimserver when building the Nix package

### LINKED ISSUE #595: Nixos: configure R.nvim for read-only plugin directories
I am using R.nvim on NixOS through a Nix-managed Neovim setup, specifically nixcats.
In this setup, plugins are installed into the Nix store, which is immutable/read-only.
When opening an `.R` file, R.nvim tries to build runtime artifacts inside its own plugin checkout. Since the checkout is located under `/nix/store`, this fails with a read-only filesystem error.

```
  Error making rnvimserver [2].
  stdout:
  gcc -pthread -O2 -Wall complete.c resolve.c hover.c definition.c signature.c rhelp.c chunk.c roxygen.c data_structures.c logging.c
  rnvimserver.c obbr.c tcp.c utilities.c ../nvimcom/src/common.c -o rnvimserver

  stderr:
  /nix/store/...-binutils-2.46/bin/ld.bfd: cannot open output file rnvimserver: Read-only file system
  collect2: error: ld returned 1 exit status
  make: *** [Makefile:17: rnvimserver] Error 1
```

I also see related errors when R.nvim tries to install/build nvimcom from the plugin directory:

```
  Why build nvimcom: Nvimcom not installed
  Warning in dir.create(path, showWarnings = TRUE, recursive = TRUE, ...) :
    cannot create dir 'nvimcom/man', reason 'Permission denied'
  Error in mydir.create(name) : failed to create directory ‘nvimcom/man’
```

  ## Environment

  - OS: NixOS
  - Neovim setup: nixcats
  - Plugin installation model: plugins are stored in /nix/store
  - R.nvim source path: read-only Nix store path
  - R and build tools are available through Nix

## Request/Issue
Would it be possible for R.nvim to support a separate writable build/state directory, independent of the immutable plugin source directory?
For example, something like:
```
  require("r").setup({
    rnvim_home = "/nix/store/.../R.nvim",      -- immutable plugin source
    build_dir = "~/.cache/R.nvim/build",      -- writable generated files
    nvimcom_lib = "~/.local/share/R/library", -- optional/user-controlled
  })
```

From looking at the code, config.rnvim_home is derived from the loaded plugin path and is then used for both reading source files and writing build artifacts.
At the moment, R.nvim appears to treat the plugin source directory (rnvim_home) as both the immutable source root and the writable build/state root.
Splitting these would make packaging much cleaner.

Alternatively, is there already a supported way to configure R.nvim for read-only plugin directories that I missed?

## Current workaround
The workaround I am using is to copy the R.nvim checkout from the Nix store into a writable cache directory and prepend that copy to Neovim’s runtimepath. Conceptually:
```
  r_root="$XDG_CACHE_HOME/nvim-bootstrap/r.nvim"
  cp -a "$nix_store_rnvim_source/." "$r_root/"
  chmod -R u+rwX "$r_root"
  nvim --cmd "set runtimepath^=$r_root"
```
This works because R.nvim then sees config.rnvim_home as the writable cache copy rather than the immutable Nix store path.

## Related Issue(s)
This is related to #229, but I believe it is a more specific remaining issue.
#229 discusses NixOS support, runtime dependencies, and packaging `nvimcom`. In my setup I can provide `gcc`, `make`, `tar`, `treesitter`, and even build `nvimcom` through Nix. However, R.nvim still tries to build `rnvimserver` and other generated artifacts inside  `config.rnvim_home`, which points to the plugin checkout under `/nix/store`.
Because `/nix/store` is read-only, this fails even when all dependencies are available.

Neurarian:
Hi,

This at least partially fixes #595.

This PR resolves the `rnvimserver` part of the issue: the binary is now compiled at build time, instead of relying on Neovim to compile it at runtime (which doesn't work well on NixOS, since the Nix store is read-only).

The package can then be consumed via, e.g.:

```nix
inputs.plugins-rNvim.packages.${pkgs.stdenv.hostPlatform.system}.default
```

with `rnvimserver` already baked in.

`nvimcom` still requires a bit of extra tinkering, but I could open a separate PR to also expose `nvimcom` as its own Nix package for easier downstream integration.

All the best,
Neurarian

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -85,6 +85,12 @@
           pname = "R.nvim";
           version = "0.1.0";
           src = ./.;
+          nativeBuildInputs = [pkgs.gnumake pkgs.gcc];
+          buildPhase = ''
+            runHook preBuild
+            make -C rnvimserver
+            runHook postBuild
+          '';
         };
       }
     );


PMassicotte:
Thank you for your PR, I will have a check later today. Cheers.

PMassicotte:

PR #617, opened 2026-09-03
fix(rnoweb): make go to previous chunk work

urtzienriquez:
In Rnoweb files, RPreviousRChunk does not navigate to the previous chunk. This PR is an attempt to fix that

### DIFF
--- a/lua/r/rnw.lua
+++ b/lua/r/rnw.lua
@@ -197,12 +197,12 @@ end
 ---@return boolean
 local go_to_previous = function()
     local curline = vim.api.nvim_win_get_cursor(0)[1]
-    local lang = get_lang()
-    if lang ~= "r" and lang ~= "python" then
-        local i = vim.fn.search("^<<.*$", "bnW")
-        if i ~= 0 then vim.api.nvim_win_set_cursor(0, { i - 1, 0 }) end
-    end
     local i = vim.fn.search("^<<.*$", "bnW")
+    local lang = get_lang()
+    if lang == "r" or lang == "python" or lang == "chunk_end" then
+        if i ~= 0 then vim.api.nvim_win_set_cursor(0, { i - 1, 0 }) end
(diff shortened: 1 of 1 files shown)


lua/r/rnw.lua
```diff
@@ -197,12 +197,12 @@ end
 ---@return boolean
 local go_to_previous = function()
     local curline = vim.api.nvim_win_get_cursor(0)[1]
+    local i = vim.fn.search("^<<.*$", "bnW")
     local lang = get_lang()
-    if lang ~= "r" and lang ~= "python" then
-        local i = vim.fn.search("^<<.*$", "bnW")
+    if lang == "r" or lang == "python" or lang == "chunk_end" then
         if i ~= 0 then vim.api.nvim_win_set_cursor(0, { i - 1, 0 }) end
```
PMassicotte: Hi and thank you for your contribution. Would you mind to add this small change to prevent crash when the chunk is at the header of the file?

```diff
diff --git a/lua/r/rnw.lua b/lua/r/rnw.lua
index 1fbf21e..9397355 100644
--- a/lua/r/rnw.lua
+++ b/lua/r/rnw.lua
@@ -200,7 +200,7 @@ local go_to_previous = function()
     local i = vim.fn.search("^<<.*$", "bnW")
     local lang = get_lang()
     if lang == "r" or lang == "python" or lang == "chunk_end" then
-        if i ~= 0 then vim.api.nvim_win_set_cursor(0, { i - 1, 0 }) end
+        if i > 1 then vim.api.nvim_win_set_cursor(0, { i - 1, 0 }) end
         i = vim.fn.search("^<<.*$", "bnW")
     end
     if i == 0 then
```

PMassicotte (changes requested):

lua/r/rnw.lua
urtzienriquez: Done. Thanks a lot for the fix!

PR #620, opened 2026-09-12
fix(send): fix RSendChain misdetecting pipe position in quarto/rmd files

PMassicotte:
I came around an issue with a long qmd document where extracted code chunk postions would become drifted by some lines in the temporary buffer we are creating to capture chunks. This caused an issue with `RSendChain` thinking that the cursor was not in a piped expression althought it was. This would give the folowing message:

```
The cursor is not inside a piped expression.
```

I found that `create_r_buffer()` used `get_node_text()` to copy R chunk content into the scratch buffer, but its handling of the trailing newline is inconsistent across Neovim versions, silently adding one extra blank line per chunk and misaligning row numbers on long documents. With some test, I found that `nvim_buf_get_lines()` is a better cross-version approach.

### DIFF
--- a/lua/r/buffer.lua
+++ b/lua/r/buffer.lua
@@ -30,40 +30,36 @@ M.create_r_buffer = function()
          ]]
     )
 
-    local contents = {}
+    local lines = {}
     local last_end = 0
 
     local root = get_root_node(bufnr)
     if not root then return end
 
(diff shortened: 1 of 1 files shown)


jalvesaq:
I haven't had this problem yet. I'm using the latest released version of nvim (0.12.5). This branch is also working well in a quick test.

jalvesaq:
Merged now! Thank you!

PR #621, opened 2026-09-19
Avoid error when an empty data.frame is created

jalvesaq:


jalvesaq:
R code to replicate the bug fixed by this PR:

```r
d <- data.frame()
```

PMassicotte (approved):

PMassicotte:
Confirmer, fix the issue on my side. Thank you very much!

PR #622, opened 2026-09-19
Object Browser: show package title instead of description

jalvesaq:


jalvesaq:
Replace long descriptions with the usually shorter title in the "libraries" view of the Object Browser.

PMassicotte (approved):

PMassicotte:
Thank you very much, works on my side and it is cleaner/nicer view.

